PR #1, opened 2015-09-24
Add landice regions

stephenprice:


stephenprice:
Initial commit of regions for Antarctic and Greenland ice sheets as defined by drainage basins (according to IceSAT and IMBIE definitions). Tags were also added to both Greenland and Antarctic features that will be useful for defining a number of other standard features (e.g., drainage basins flowing into the Ross and Filchner-Ronne ice shelves, basins making up East vs. West Antarctica, etc.).

@douglasjacobsen - I'm flagging you on this for now since Matt's name does not seem to show up with "@". Feel free to forward to him as satisfying this one should be pretty straightforward.

matthewhoffman:
Seems fine to me.  Thanks @stephenprice for getting these set up!

PR #3, opened 2016-01-20
Adds requirement for plot_features.py to README

pwolfram:
Fixes README and reduces ambiguity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,14 @@ this repository. All of them can be run with the -h flag to get usage
 information.
 
 A typical workflow will look like:
-	* Build feature list (features.geojson):
-		- Multiple calls to: ./merge_features.py
-	* Edit features:
-		- Edit features.geojson
-	* Visualize features:
-		- ./plot_features.py (Note the requirements to use this listed in its usage information)
-	* Split features:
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, looks good.  It doesn't look like there's any easy way to make `plot_features.py` give its help message if `basemap` isn't present so I think this solution is fine.

douglasjacobsen:
An easy way to catch if basemap isn't present is the following:

```
try:
    from mpl_toolkits.basemap import Basemap, addcyclic
except ImportError as e:
    print "ERROR: Cannot find either basemap or addcyclic. Error message follows:"
    print e
    sys.exit(1)
```

xylar:
@douglasjacobsen, that would work but I think that's not really any more helpful than the default error message:
```
  File "./plot_features.py", line 19, in <module>
    from mpl_toolkits.basemap import Basemap, addcyclic
ImportError: No module named basemap
```
The main thing is not to tell people that they can find out the requirements by loading the help message, and then they find out that fails because they're missing the requirement.  This way, you don't expect to get that info from the help message, so it's okay.

douglasjacobsen:
OK, sounds good.

PR #4, opened 2016-01-21
Fix bug in subroutine for comparing tags (match_tag_list)

xylar:
Previously, match_tag_list would always return False for multiple input tags in the previous version.  Now, match_tag_list returns true if the any of the requested tags are present in a given feature.

### DIFF
--- a/utils/feature_test_utils.py
+++ b/utils/feature_test_utils.py
@@ -9,12 +9,11 @@ def match_tag_list(feature, master_tags):#{{{
 	
 	feature_tag_list = feature_tags.split(';')
 
-	test = True
+	test = False
 	for tag in master_tags:
 		if tag in feature_tag_list:
-			test = test and True
-		else:
-			test = test and False
+			test = True
+			break
 
 	return test
 #}}}


xylar:
### Testing ###
I needed a feature file with all the land-ice basins from Antarctica.  The following works with the new code but not with the old:

```
./merge_features.py -t "WestAntarcticaIMBIE;EastAntarctiaIMBIE" -d landice/region/
```

douglasjacobsen:
@xylar The tag list is intended to match *all* of the tags, not just one of them.

If the tag list matches any tag, then it's not possible to build an intersection of tags.

douglasjacobsen:
@xylar Based on your last comment, the old functionality would work if you did this:

```
./merge_features.py -t "WestAntarcticaIMBIE" -d landice/region/
./merge_features.py -t "EastAntarctiaIMBIE" -d landice/region/
```

xylar:
Okay, thanks @douglasjacobsen.  I must have misunderstood the documentation.

PR #9, opened 2016-01-31
Add a script for splitting features at +/- 180 degrees (the antimeridian)

xylar:
The script uses the 'shapely' package to cut each feature in a feature collection with a very thin wedge around the antimeridian (+/- 180 degrees longitude).  Features that do not intersect the wedge are left unchanged.

Also three land-ice basins are update, two based on splitting at the antimeridian and the formatting of a third is corrected.


xylar:
I tested this branch by applying it to the Antarctic land-ice basins, to the ocean regions and to the ice-shelf regions (soon to be committed).  The land-ice basins that are split using this script are included in this pull request.  The only ocean region that is affected by this script would be the Arctic because there is not a point at the pole (a requirement of this script).  I have chosen not to modify the Arctic basin to meet the requirements of this PR,but would be willing to do so if anyone thinks it is necessary.

xylar:
All ocean basins except the Arctic pass through the `fix_features_at_antimeridian.py` script unchanged.  A side effect of how the script detects intersections with the antimeridian is that it requires there to be a point at the pole for regions that contain the pole.  Since this is not the case for the Arctic region as currently defined, the tool attempts to add the point at the pole. While this would not be a problem, it is also not necessary so I haven't modified the Arctic region.

PR #10, opened 2016-01-31
Add ice shelf regions

xylar:
Regions covering 100 ice shelves are added.  Where ice shelves intersect multiple IMBIE basins, they have been divided into numbered subregions.  All features are tagged with the ice-shelf name, the IMBIE basin name, and the IMBIE region name.

Each region covers not just the present-day location of the shelf but also all parts of a given basin that are closer to that shelf than any other shelf in the basin.  The regions extend to the continental-shelf break as defined by the -800 m contour.  Thus, the regions should be useful as shelves advance or retreat over time.


xylar:
The ice-shelf regions that cross the antimeridian have been divided into multiple polygons using the fix_features_crossing_antimeridian.py script from #9.

stephenprice:
Looks great Xylar. I'll try to check this out and merge later today. Thanks for all your work on this (and the new tool as well).

stephenprice:
@xylar  - I spot  checked a few of the more standard ones I'm familiar with here and they all look OK to me. 

@douglasjacobsen - I'm not sure what the process is for merging these, since we don't really have any obvious way of 'testing' them out. I'm happy to check out a branch and merge the appropriate way, but let me know if this is also a case where we can (horrors!) just hit the green button.

douglasjacobsen:
@stephenprice If you're happy with the way they look (you can just plot them in github, or using the plot script) that's good enough. Alternatively you could plot them using the mask creator, and see how well they work.

stephenprice:
@douglasjacobsen - I checked them out w/ the github plotting script. That's all I'm technically capable of doing right now in terms of testing. If that is adequate, I will merge. If not, I can wait until I have some other means of looking through them and testing them out.

douglasjacobsen:
@stephenprice You're free to merge them once you feel confident that they are good enough. So, if you're confident that the plotting you've done so far is good enough, feel free to merge them.

PR #12, opened 2016-02-05
Additional projection types

pwolfram:
This PR cleans up the plotting routine and adds additional projection types

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -23,6 +23,26 @@ def plot_base(maptype): #{{{
         map = Basemap(projection='ortho', lat_0=45, lon_0=-100, resolution='l')
         map.drawmeridians(np.arange(0,360,30))
         map.drawparallels(np.arange(-90,90,30))
+    elif maptype == 'aeqd':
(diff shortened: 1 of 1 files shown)


pwolfram:
@xylar, here is the PR with some additional projections you can try out.

xylar:
@pwolfram, thanks for looking into this.  I'll fulfill the PR as soon as I can find the time to do a proper test.

xylar:
@pwolfram, thanks for these additional projections.  I think they will be useful.

I verified that all 5 new produce plots of features that look reasonable.  Here is an image with 4 of them with that Arctic and Southern Ocean:
test
I found a fix for the Arctic basin (first projecting the lat/lon coordinates to map coordinates, then plotting, rather than using the latlon flag), which I used here and I will make a PR for that change momentarily.

xylar:
@pwolfram, I don't seem to have permission to delete your remote, so you will need to do that yourself.

pwolfram:
Done, thanks @xylar!

PR #15, opened 2016-02-07
Modify plotting script to remove out-of-bounds artifacts

xylar:
Several projections were producing artifacts when shapes were out of bounds of the projection.  This merge fixes this issue by detecting out-of-bounds points and setting them to NaN.

This merge makes several other enhancements that improve speed and functionality of the plotting script:
* Plotting has been reorganized so one Basemap object is created per projection, not one per projection per feature
* Edges of features that are longer than 1 degree are broken into a series of 1-degree segments for better plotting
* For projections that handle them correctly, features are plotted as translucent polygons as well as outlines
* Each projection is saved to a separate file with the projection name appended to the end of the file name. Four subplots in a single figure was typically too small to be useful.  Also, it may be more useful to have more (or less) than 4 plots. By default, all 17 projections are included.
* A new command-line option -m has been added to optionally provide a list of comma-separated map types.  If none is provided, all are used.

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -5,18 +5,25 @@ matplotlib's basemap.
 
 It requires basemap: http://matplotlib.org/basemap/
 
-The -r flag is used to pass in a features file that will be plotted, and the -o
+The -f flag is used to pass in a features file that will be plotted, and the -o
 flag can optionally be used to specify the name of the image that will be
-generated.
+generated.  If more than one map type is used, the name of the map type will
+be appended to the image name.  The option -m flag can be used to specify a 
+comma-separated list of map types to be plotted.  If no map type is specified, 
+all maps are used.  Possible map types are 'ortho', 'aeqd', 'eck4', 'cyl',
+'merc', 'vandg', 'mill', 'mill2', 'robin', 'robin2', 'hammer', 'northpole', 
+'southpole', 'atlantic', 'pacific', 'americas', 'asia'
 
-Author: Phillip J. Wolfram
-Date: 08/25/2015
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, I think I've done the best that can be done without some really fancy stuff that knows about the boundaries of each projection.  I added support for translucent polygons where they don't cause trouble, but if they intersect the boundaries they cause weirdness.  The only way around that would be using something like shapely to intersect each shape with the shape of the boundary.  Way too much work for this project.

Let me know what you think.

xylar:
### Testing ###

I ran the following:

```
$ rm features.geojson
$ ./merge_features.py -d ocean
$ ./merge_features.py -f iceshelves/region/Ronne_1/region.geojson
$ ./merge_features.py -d landice
$ ./plot_features.py -f features.geojson
```
The resulting plots seem to be well behaved:
features_aeqd
features_americas
features_asia
features_atlantic
features_cyl
features_eck4
features_hammer
features_merc
features_mill
features_mill2
features_northpole
features_ortho
features_pacific
features_robin
features_robin2
features_southpole
features_vandg

pwolfram:
@xylar, thank you very much for making what was original a prototype into production-level code.  I've made a few minor changes to your code in https://github.com/xylar/geometric_features/pull/1 and am happy to make this merge assuming you are ok with those minor changes, the largest is to change the colorscheme to be a little nicer, e.g.,

features_vandg

xylar:
@pwolfram, I merged your changes in.  Feel free to merge when you're ready.

pwolfram:
@xylar, thanks.  These changes are great.  I think you'll have to delete your branch because I don't think I can do that.

PR #16, opened 2016-02-08
Change plot_features to use cartophy

xylar:
This limits the types of projections that can be used but has the very large advantage that it supports polygons for a wider variety of projections and handles boundaries correctly.

(Cartophy also does a reasonable job of subdividing lines during projection, though I left in code for subdividing segments longer than 4 degrees, which results in smoother projections.)

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -18,11 +18,10 @@ Authors: Xylar Asay-Davis, Doug Jacobsen, Phillip J. Wolfram
 Last Modified: 02/08/2016
 """
 
-import numpy as np
+import os.path
 import json
+import numpy as np
 import matplotlib.pyplot as plt
-import os.path as ospath
-
 import cartopy.crs
 import cartopy.feature
 
@@ -190,7 +189,7 @@ if __name__ == "__main__":
         mapTypes = args.map_type.split(',')
 
     if not args.features_plotname:
-        args.features_plotname = ospath.splitext(args.features_file)[0] + '.png'
+        args.features_plotname = os.path.splitext(args.features_file)[0] + '.png'
 
     projections = build_projections()
 
@@ -203,7 +202,7 @@ if __name__ == "__main__":
         if(len(mapTypes) == 1):
             plotFileName = args.features_plotname
         else:
-            plotFileName = '%s_%s.png'%(ospath.splitext(args.features_plotname)[0],mapType)
+            plotFileName = '%s_%s.png'%(os.path.splitext(args.features_plotname)[0],mapType)
         mapInfo[mapType] = (ax, projection, plotFileName, fig)
 
     plot_features_file(args.features_file, mapInfo)
--- a/plot_features.py
+++ b/plot_features.py
@@ -26,43 +26,28 @@ import os.path as ospath
 import cartopy.crs
 import cartopy.feature
 
+def build_projections(): #{{{
+    projections = {}
+    projections['cyl'] = cartopy.crs.PlateCarree()
+    projections['merc'] = cartopy.crs.Mercator()
+    projections['mill'] = cartopy.crs.Miller()
+    projections['mill2'] = cartopy.crs.Miller(central_longitude=180.)
+    projections['moll'] = cartopy.crs.Mollweide()
+    projections['moll2'] = cartopy.crs.Mollweide(central_longitude=180.)
+    projections['robin'] = cartopy.crs.Robinson()
+    projections['robin2'] = cartopy.crs.Robinson(central_longitude=180.)
 
-def plot_base(mapType): #{{{
+    projections['ortho'] = cartopy.crs.Orthographic(central_longitude=-100., central_latitude=45.)
+    projections['northpole'] = cartopy.crs.Orthographic(central_longitude=0., central_latitude=90.)
+    projections['southpole'] = cartopy.crs.Orthographic(central_longitude=0., central_latitude=-90.)
+    projections['atlantic'] = cartopy.crs.Orthographic(central_longitude=0., central_latitude=0.)
+    projections['pacific'] = cartopy.crs.Orthographic(central_longitude=180., central_latitude=0.)
+    projections['americas'] = cartopy.crs.Orthographic(central_longitude=-90., central_latitude=0.)
+    projections['asia'] = cartopy.crs.Orthographic(central_longitude=90., central_latitude=0.)
 
-    if mapType == 'cyl':
-        projection = cartopy.crs.PlateCarree()
-    elif mapType == 'merc':
-        projection = cartopy.crs.Mercator()
-    elif mapType == 'mill':
-        projection = cartopy.crs.Miller()
-    elif mapType == 'mill2':
-        projection = cartopy.crs.Miller(central_longitude=180.)
-    elif mapType == 'moll':
-        projection = cartopy.crs.Mollweide()
-    elif mapType == 'moll2':
-        projection = cartopy.crs.Mollweide(central_longitude=180.)
-    elif mapType == 'robin':
-        projection = cartopy.crs.Robinson()
-    elif mapType == 'robin2':
-        projection = cartopy.crs.Robinson(central_longitude=180.)
+    return projections #}}}
 
-    elif mapType == 'ortho':
-        projection = cartopy.crs.Orthographic(central_longitude=-100., central_latitude=45.)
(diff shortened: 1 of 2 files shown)


xylar:
@pwolfram, at @pelson's suggestion, I rewrote the plotting script with cartophy.  The results are much better.

I don't think it's important that cartophy doesn't support all the projections we were using before.  Many of them are redundant for our purposes.

One nice side effect of using cartophy is that it is smart about determining the bounds of the features you're plotting (like matplotlib) so if you are only plotting Greenland, for example, it will zoom in on it.

ocean_americas
ocean_asia
ocean_atlantic
ocean_cyl
ocean_merc
ocean_mill
ocean_mill2
ocean_moll
ocean_moll2
ocean_northpole
ocean_ortho
ocean_pacific
ocean_robin
ocean_robin2
ocean_southpole

xylar:
@pwolfram, hmm, better hold off a bit.  I just did some tests on Greenland and Antarctica.  Outlines work but filled polygons only work where there isn't land.  Presumably a z-level issue.  I'll see if there's anything I can do about it.

greenland_merc
ice_shelves

xylar:
@pwolfram, okay, I fixed it with some explicit `zorder` arguments.

greenland
ice_shelves

pwolfram:
@xylar, this is excellent.  I'm not worried about loosing projection options.  Based on the plots above the functionality required appears to be available.  Are you comfortable with me starting the review process sometime soon?

pwolfram:
I made some minor edits at https://github.com/xylar/geometric_features/pull/2 and am happy to merge when you are ready.

xylar:
@pwolfram, yes, go ahead.

pelson:
This is so cool! Incredible work @xylar! If you wanted to submit a gallery addition to cartopy, that would be a great opportunity to provide a cross-link to this repo :wink: and I'd love to see at least one of these plots there! http://scitools.org.uk/cartopy/docs/latest/gallery.html

plot_features.py
```diff
@@ -1,122 +1,98 @@
...
-        map.drawparallels(np.arange(-90,90,30))
+import os.path as ospath
+
+import cartopy.crs
+import cartopy.feature
+
+
+def plot_base(mapType): #{{{
+
+    if mapType == 'cyl':
```
pelson: This kind of thing is crying out for a dictionary...

plot_features.py
```diff
@@ -134,52 +110,39 @@ def divide_poly_segments(points): #{{{
...
-            y[mask] = np.nan
-            if(filled and supportsFill):
-                xy = zip(x,y)
-                poly = Polygon( xy, facecolor=color, alpha=0.4)
-                plt.gca().add_patch(poly)
-            map.plot(x, y, linewidth=2.0, color=color)
+        (mapType, ax, projection, plotFileName, fig) = mapInfo[mapIndex]
+        x = points[:,0]
+        y = points[:,1]
+        ax.fill(x, y, transform=refProjection, color=color, alpha=0.4, zorder=3)
```
pelson: The fill and lines should be achievable in a single call. If not, you could even use a shapely geometry to visualise.

plot_features.py
xylar: @pelson, I think you're right.  My initial thought was that we only want to build the projections we're actually going to use, but they don't seem to be expensive to build.  I'll make this change.

plot_features.py
xylar: Okay, I'll try that out.  Thanks for the tip.

xylar:
@pwolfram, I'm going to make the changes that @pelson recommends.  I'll let you know when you can have another look.

xylar:
@pwolfram, take a look at the latest changes.  Should now be ready to merge.

xylar:
@pelson, I will think of how best to do a submission to the cartopy gallery.  The trouble is that we store our features as a large number of individual geojson files.  These have to be pre-processed into a single file for plotting.  We could more easily do an example that downloads and plots a single one of our regions.

Anyway, I'll think about it and get back to you.  If you happen to be more interested in sharing an image without an accompanying script for generating the image, that would be no problem.

pelson:
> If you happen to be more interested in sharing an image without an accompanying script for generating the image, that would be no problem.

I'd be happy enough to put some docs which explain the steps (and the source) of the data, without requiring the gallery example to do that directly.

pwolfram:
@xylar, I think I've come to a resolution about the cleanest way to do imports (my biggest comment).  You can see my changes in this direction at https://github.com/xylar/geometric_features/pull/3.  I agree with @pelson, the plots are really nice!

pwolfram:
@pelson, as @xylar noted on a previous comment the figures can be built with commands like

```
$ rm features.geojson
$ ./merge_features.py -d ocean
$ ./merge_features.py -f iceshelves/region/Ronne_1/region.geojson
$ ./merge_features.py -d landice
$ ./plot_features.py -f features.geojson
```

from the soon to be merged master of https://github.com/MPAS-Dev/geometric_features.

Essentially, `merge_features.py` builds up a geojson file that can then be plotted with `plot_features.py`.  A variety of useful regions for climate science are built up in the `iceshelves`,  `landice`, and `ocean` directories. The user has the option to choose what regions are to be included in the plot by following the workflow above.  The `-d` option allows a whole directory (category really) to be included and `-f` to alow only a single region geojson file to be included.

What else do you need beyond this to include these great plots by @xylar in the cartopy gallery?

xylar:
@pelson, I would probably recommend using an image of the Greenland and/or Antarctic basins.  They are the least chaotic to visualize, though there's a little more work required to build up the features.  Here's an example bash script:
``` bash
#!/bin/bash
rm features.geojson
for file in landice/region/Antarctica*/*.geojson
do
  ./merge_features.py -f $file
done
./plot_features.py -f features.geojson -o AntarcticBasins.png -m southpole

rm features.geojson
for file in landice/region/*Greenland*/*.geojson
do
  ./merge_features.py -f $file
done
./plot_features.py -f features.geojson -o GreenlandBasins.png -m merc
```

Here are the results:
antarcticbasins
greenlandbasins

pwolfram:
@xylar, a note about one minor potential bug I saw:

The Arctic does not appear to be plotted correctly for features_moll.png (it should be brown, not background white):

screenshot 2016-02-09 09 28 31

However, it is ok for other views, e.g., 

screenshot 2016-02-09 09 29 42

Thus, there are still some minor issues with ensuring each region is plotted correctly for each projection.  

However, we are a magnitude better than we were before.  I'm going to merge this with the recognition that some projections and regions may not be 100% compatible.  However, everyone probably agrees that this is ok because one wouldn't expect to see the Arctic in a south pole plot, for example.

pelson:
> The Arctic does not appear to be plotted correctly for features_moll.png (it should be brown, not background white):

Interesting. There is a "threshold" in cartopy which identifies geometries which are sufficiently close to the boundary (Proj.4 gives no such information, so it has to be a heuristic). It may be that in this case, the Mollweide projection's threshold is simply too high. I gave a StackOverflow answer a few days ago which gave a hacky way of testing whether a higher threshold might be helpful: http://stackoverflow.com/questions/34898312/cartopy-background-color-outside-of-data-domain/34989375#34989375. The magic numbers in cartopy are just that at the moment, they have only been determined with very little investigation, and can most certainly be adjusted at this point (we have plans for making the threshold less and less important in the future).

xylar:
@pelson, I tried your suggestion.  I used a threshold of `1e4` and then `1e3` but the poles still seem to give the Mollweide plot the same difficulty.
```
class BetterMollweide(cartopy.crs.Mollweide):
    def __init__(self, *args, **kwargs):
        cartopy.crs.Mollweide.__init__(self, *args, **kwargs)
        self._threshold = 1e3

    @property
    def threshold(self):
        return self._threshold
```

@pwolfram, I think this is not something we are going to worry about on our side.  MollWeide isn't a very good projection for looking at regions near the poles anyway, so it isn't likely to matter that it's not currently supported.

pelson:
Hmmm. Thanks for trying @xylar. Looks like there must be a deeper problem. If you have the geometry handy, and can put it into a SSCCE on the cartopy issue tracker, I'll try to dig a little deeper.

xylar:
@pelson, as you no doubt have seen, I submitted https://github.com/SciTools/cartopy/issues/724 to the cartopy issue tracker.  I'm linking to it here so we know it's related to this pull request.

PR #18, opened 2016-02-10
Update the Southern Ocean to be between 90 S and 40 S latitude

xylar:
This merge adds a script, `difference_features.py`, to mask out (difference) features.  One or more features are supplied with the -f flag and one or more masking features are supplied with -m.  The result is the first set of features with the intersection with the second set of features removed (i.e. the difference as defined by the shapely library).  The results are stored in features.geojson.

The Southern ocean domain was updated to include the full Antarctic continent so it can handle retreat of land ice and its northern limit was moved from 60 S to 40 S latitude, as decided by the ACME ocean modeling team.

Several oceans that would have intersected the new limits of the Southern Ocean have been clipped to lie above 40 S.  Features were modified as follows:
```bash
#!/bin/bash

for feature in South_Atlantic_Ocean South_Pacific_Ocean Indian_Ocean Tasman_Sea Bass_Strait Great_Australian_Bight
do
  ./merge_features.py -f ocean/region/$feature/region.geojson
done
mv features.geojson intersect_southern_ocean.geojson
./difference_features.py -f intersect_southern_ocean.geojson -m ocean/region/Southern_Ocean/region.geojson
./split_features.py -f features.geojson
'''

### CODE REVIEW

**Comment by xylar** [comment]
@douglasjacobsen, feel free to assign someone else if you'd rather not review this.

**Comment by xylar** [comment]
@milenaveneziani, please take a look at these files and make sure they are consistent with the changes you were wanting to the northern boundary of the Southern Ocean.  I realize this is only the first part of several changes/enhancements that we want to make to these regions.

modified

**Comment by milenaveneziani** [comment]
Thank you @xylar  for doing this. Checking the files asap.

**Comment by douglasjacobsen** [comment]
@xylar Thanks for this script, it should make chopping regions a lot easier.

**Comment by douglasjacobsen** [comment]
Also, as a small note (complete aside) we might want to pick specifications for how to format python scripts in this repo. I'm not going to worry about it right now, but your script has both spaces and tabs, and we might want to clean it up at some point (and maybe make all of them consistent one way or the other).

**Comment by xylar** [comment]
@douglasjacobsen, I completely agree that we should have a standard.  The python best practices document that @pwolfram shared (I don't have it handy) recommends spaces over tabs, and that is my strong preference.  (Sorry for missing a few tabs -- I was working from one of your scripts that used tabs and must have missed some.)

**Comment by pwolfram** [comment]
@xylar and @douglasjacobsen, I recommend using modelines to enforce this in vim: https://github.com/MPAS-Dev/geometric_features/blob/master/plot_features.py#L210

Then, you can just highlight everything and use "=" to apply the formatting.

**Comment by xylar** [comment]
@pwolfram, when I tried that, it replaced all my spaces with the tabs I so much despise.  Do I have something set wrong?

**Comment by pwolfram** [comment]
So it worked for me because my tabs turn into triangles and I only have spaces.  I'm suspicious it has something to do with my vimrc:  https://gist.github.com/d646c8398c509a8d9e67

perhaps it is one of these lines

```
145 set tabstop=2
146 set shiftwidth=2
147 set softtabstop=2
148 set expandtab
```

The line `# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python` means automatic indent, tab spaces = 4, soft tab spaces=4, expand tabs, shift width=4, format=python

Hence, I'm a little surprised it isn't working for you but am suspicious it is in your vimrc settings.

@douglasjacobsen may have a better feel for this because he originally proposed the modelines.

milenaveneziani:
@xylar, @douglasjacobsen: I was just checking the files/new regions, and was going to comment on a few things:
1) I would not change the original regional areas, even though the new SO cuts through them. This is the case for 3 regions: Great Austrialian Bight, Tasman Sea, and Bass Strait. Their definition should be independent from how the SO is defined.
2) For the sake of redefining the Indian and Pacific Ocean, I would instead follow this workflow:
a) cp the region.geojsan files for Great Austrialian Bight, Tasman Sea, Bass Strait, Indian and South Pacific Ocean to a *_tmp version (say Great_Austrial_Bight_tmp/region.geojsan, etc);
b) reset Bass_Strait/region.geojson, Great_Australian_Bight/region.geojsan, and  Tasman_Sea/region.geojsan to their original versions (pre-PR);
c) define new Indian Ocean as union of: Indian_Ocean_tmp, Bass_Strait_tmp, Great_Australian_Bight_tmp, Mozambique_Channel, Gulf_of_Aden, Gulf_of_Oman, Arabian_Sea, Laccadive_Sea, Bay_of_Bengal, Andaman_or_Burma_Sea (west of Bay of Bengal), Malacca_Strait (I think I have included them all..);
d) define new South Pacific as union of: South_Pacific_tmp, Tasman_Sea_tmp, Coral_Sea, Solomon_Sea and Bismarck_Sea;
e) leave South_Atlantic as is.

Makes sense?

Then the real hurdle will be redefining the whole Pacific and Atlantic, with all those marginal seas to account for. But I can help with that. I will start making a list of all the subregions I think they should include, double checking at the end than the total sum will be the same as the original sum of regions..

douglasjacobsen:
@xylar and @pwolfram, Even though I hate spaces I'm fine with using them in this repo for python.

However, even after adding the modeline, my `=` doesn't auto format correctly (it does change tabs to spaces, but it adds weird indentations that make it invalid python). Modelines really just make it so that in the future the default for each python file looks similar, but I don't think they help you make the file standardized using the autoformatting.

@xylar You can use `:%s/\t/    /g` to replace tabs with 4 spaces in the file. Then the modeline will auto convert tabs you insert into spaces for you.

xylar:
@douglasjacobsen and @pwolfram, I use spyder rather than vim to edit my python files -- it's much more convenient for me.  It can handle tabs instead of spaces and 4 spaces instead of my default of 2, but it doesn't change on the fly (I have to go into settings to change it every time).  I can easily search and replace for tabs, but I won't notice them typically because they're (I think) 3 or 4 spaces so they look reasonable, unlike when they're 8 spaces in vim.

douglasjacobsen:
Yeah, and modelines are not going to help you at all. I can go through and fix the existing scripts and add modelines to them, so then in the future you shouldn't have to worry about replacing any of them.

pwolfram:
@douglasjacobsen, if you would like I can fix the existing scripts because my autocorrect, at least the last time I checked, works.

PR #20, opened 2016-02-10
Added tags for 6 major ocean basins to ocean regions

xylar:
Nearly all ocean regions should now belong to one of: Pacific, Atlantic, Arctic or Indian Oceans

Also, this merge adds a script, `tag_features.py`  for adding a tag to a feature collection.

PR #21, opened 2016-02-11
Add a mask for the MOC region

xylar:
This merge also modifies the difference_features.py script to handle cases where 1) a feature does not interesect the mask at all (so nothing should happen) and 2) a feature intersects the mask completely so it should be removed entirely.


xylar:
@milenaveneziani, this should be merged only after #20.

xylar:
### Workflow for making MOC basins ###

To make an MOC basin from region that have been tagged appropriately, the workflow is as follows:
```bash
rm features.geojson
./merge_features.py -d ocean -t Atlantic_Basin
mv features.geojson Atlantic_Basin.geojson
./difference_features.py -f Atlantic_Basin.geojson -m ocean/region/MOC_mask/region.geojson
mv features.geojson Atlantic_MOC.geojson
./plot_features.py -f Atlantic_MOC.geojson -o Atlantic_MOC.png -m cyl
```

The same should work with `Atlantic ---> Pacific` and `Atlantic ---> Indian`.  Here are the results:

atlantic_moc
indian_moc
pacific_moc

xylar:
@douglasjacobsen, @pwolfram, @stephenprice, @milenaveneziani  and anyone else with an opinion, do we want to have a place in this repository (or elsewhere) for storing scripts like `make_MOC_basin.bash` above to make certain operations easier?  For now, I'm just including them in the PRs and assuming others will figure out their own workflow.

pwolfram:
@xylar, I think any scripts in the workflow should be under version control.  At a minimum this allows them to be improved over time and it helps keep a record of the workflow at a given commit (point in time).  The scripts represent the high-level thought process and may be some, if not the most, important part of the code base which is really a description of a logical workflow.  The scripts operate on just this repo so they should be stored in this repo.

douglasjacobsen:
I think as long as the scripts aren't bash or csh it's fine. We've have issues with supporting bash / csh scripts in the past so I've kind of taken the stance that we should use environment agnostic languages as much as possible.

xylar:
Okay, I can "pythonize" this script.  My experience is that it messes up the `.merge_features.py` to put anything under `ocean` than isn't a geojson file.  So I guess I'd create a `scripts/ocean` folder and put it there?

douglasjacobsen:
I can look into fixing the merge_features.py issue.

xylar:
@douglasjacobsen, okay, so supposing `merge_features.py` is smart enough to just use geojson files, would you prefer to have scripts specific to the ocean in `ocean/scripts` rather than `scripts/ocean`?  That would seem a bit more practical.

douglasjacobsen:
I'm unsure if we need ocean in the name if it's `scripts/ocean`. It might be nice to just have something like `driver_scripts` with these assorted scripts for performing standard actions in it.

That way if the sea ice model wants to use the same regions they can just look in there and see "setup_arctic_basins.py" or something similar to that.

douglasjacobsen:
@xylar: https://github.com/MPAS-Dev/geometric_features/pull/22 should fix the issue in merge_features.py. Feel free to merge / review whenever you have time / interest.

xylar:
@milenaveneziani, okay I added a script to generate these 3 basins automatically.  Just run:
```bash
./driver_scripts/setup_MOC_basins.py
```
You should get the basins for each ocean (Atlantic, Pacific and Indian) and the MOC basins for each, cropped to be above 30 S.

Feel free to merge this branch once #20 is merged and once you're happy with these changes.

milenaveneziani:
Looks good to me. I'd just double check with Mat about the ocean points on the eastern (western) boundary of the Indian (Pacific) basins. will do that tomorrow.

xylar:
@milenaveneziani, I moved some bug fixes out of this PR.  That means that you should wait until #25 is merged before testing this PR.  Without those fixes, you will get errors when you run the driver script.

milenaveneziani:
@xylar : about the connections between Indian and Pacific Oceans, it is indeed important that there are none that are relevant in terms of mass transport. We therefore need to at least exclude the Indonesian Throughflow, i.e. have a southern boundary for Indian and Pacific at 6S.
Here are the actual options we have:
1) We only have two basins for this group: Atlantic and IndoPacific, as we talked about in the beginning (southern boundary at 30S).
2) We add to the 2 regions in 1) two more basins: Indian and Pacific with southern boundary at 6S.
The priority is to do 1) now. I guess we can do 2) later if necessary. How does that sound?

xylar:
@milenaveneziani, It's easy enough for me to make basins for both options 1) and 2) above.  I will update the drive script to produce an IndoPacific basin that is the union of the Indian and Pacific basins as current computed.  I'll also add a mask that removes everything below 6 S and create Indian (very small) and Pacific basins with this as the new southern boundary.

milenaveneziani:
@xylar: if it's easy/quick enough to do, then thanks for going ahead with both of these. But if you find that it takes longer that you had imagined, then 1) will definitely suffice for now.
Thanks again.

xylar:
@milenaveneziani, okay, here are the new basins, i think as you requested.

atlantic_moc
indian_moc
indopacific_moc
pacific_moc

If you want to go ahead and check this PR, you will just need to download the fixed version of difference_features.py from here:
https://raw.githubusercontent.com/xylar/geometric_features/666f3a1c9be632253acca01eef32ffc0c3edb296/difference_features.py
Without that fix, the driver script won't work.

Let me know how it goes.

milenaveneziani:
@xylar : I made all the relevant plots with the right script you pointed me to, and I only realized one thing: I believe the Atlantiv_MOC basin should have the Mediterranean in it, so that we make sure we include the Mediterranean outflow influence on MOC/MHT calculations.
Thanks, and I think this is really the last modification here..

xylar:
@milenaveneziani, done:
atlantic_moc

milenaveneziani:
wasn't expecting a reply today, let alone a fix.. Thanks @xylar.

I tested this and it worked after having to add the shapely package and then upgrading cartopy (because the shapely version broke a feature of the cartopy version I had). This had nothing to do with set_MOC_basins, but with the plot_features.py routine.

Another thing I noticed is that setup_MOC_basins didn't work on my laptop, for some reason. It gave me this error:
 * merging features to make Atlantic Basin
 * masking out features south of MOC region
  File "./driver_scripts/setup_MOC_basins.py", line 39, in <module>
    subprocess.check_call(args, env=os.environ.copy())
  File "/Applications/anaconda/lib/python2.7/subprocess.py", line 535, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/Applications/anaconda/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/Applications/anaconda/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/Applications/anaconda/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 13] Permission denied

So, I created the geojson files on mustang, then transferred them to my laptop and plotted the results (because I cannot use cartopy as is now on IC machines).

After all these PR's are fulfilled I will try running everything again on my laptop, to make sure I can create and plot all the basins in one place.

xylar:
@milenaveneziani, my guess is that you downloaded the difference_features.py script onto your laptop as I suggested but that it didn't have execute permission.  (You would have needed to do a `chmod u+x difference_features.py`.)  I'm not sure why it *did* work on IC.

I'll try to put the files that cartopy needs somewhere on IC and I'll post instructions on how to get it working once I do.

milenaveneziani:
@xylar you are right. Once I added the right permission to the file I downloaded, it worked fine (and it worked on IC in the first place because for some reason the permissions were right over there).

PR #22, opened 2016-02-11
Update merge_features.py to only recognize geojson files

douglasjacobsen:
This merge updates the merge_features.py script to only recognize
geojson files, which allows other files to exist within the ocean
directory and this script will ignore them.


xylar:
Looks good and does exactly what I wanted!

PR #25, opened 2016-02-12
Fix several issues with difference_features.py

xylar:
This merge modifies the difference_features.py script to handle cases where:

1. a feature does not intersect the mask at all (so nothing should happen)
2. a feature intersects the mask completely so it should be removed entirely.

Features already in features.geojson are now retained but the mask is not applied to these features, only to those being added.

Also, a missing `import sys` is added to a utils script.

### DIFF
--- a/difference_features.py
+++ b/difference_features.py
@@ -45,13 +45,14 @@ if os.path.exists(out_file_name):
     except:
         pass
 
+featuresToMask = defaultdict(list)
+
 try:
     with open(args.feature_file) as f:
         feature_file = json.load(f)
 
     for feature in feature_file['features']:
-        if not feature_already_exists(features, feature):
-            features['features'].append(feature)
+        featuresToMask['features'].append(feature)
 
     del feature_file
 except:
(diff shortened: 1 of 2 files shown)


xylar:
@douglasjacobsen, I pulled these changes out of #21 so we can keep PRs that change scripts separate from those that modify regions, as you requested.  I'm still keeping the driver script for generating MOC basins in #21 because I think the region and the driver script belong together in that case. But these bug fixes are not directly related to #21 even though they are needed to make MOC basins.

douglasjacobsen:
@xylar OK thanks. I think it's fine for driver scripts to be added with the other PR, since they aren't useful outside of that context.

milenaveneziani:
@douglasjacobsen , @xylar : should this PR be closed then, or can it be fulfilled as is for this one time? thanks.

douglasjacobsen:
@milenaveneziani I'll merge this PR as soon as I get a chance to review it. But this one is fine. #21 can be merged independently of this one now though.

xylar:
@milenaveneziani and @douglasjacobsen, #21 won't work correctly without this fix, so it really needs to wait on this merge.  Anyway, I have some work to do on #21 before it is ready to merge...

douglasjacobsen:
@xylar OK, thanks for the clarification. I'll try to get around to this tomorrow.

xylar:
@douglasjacobsen, no rush.  I just suggested that @milenaveneziani just manually copy over the file for purposes of testing #21

PR #27, opened 2016-02-25
Add script for splitting features at prime meridian

xylar:
The script splits polygons (or multi-polygons) into a west and an east half at the prime meridian (lon=0) to make the features compatible with coordinate systems that run from 0 to 360 instead of -180 to 180.

### DIFF
--- a/fix_features_at_prime_meridian.py
+++ b/fix_features_at_prime_meridian.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+
+"""
+
+This script is used to split features across the prime meridian.  The 
+features should not cross the antimeridian (+/-180 lon), so the user should
+call fix_features_at_antimeridian.py first if necessary before using this
+tool.  The script reads from a single files containing the collection of 
+features to be split at the prime meridian (the file can be 'features.geojson')
+and produces a file 'features.geojson' where all features are made up of
+polygons that do not cross the prime meridian.
+
+The script makes use of the shapely library.
+
+"""
+
+import os
+import numpy
+import json
+import argparse
+from utils.feature_write_utils import write_all_features
+
+import shapely.geometry
+import shapely.ops
+
+
+def splitGeometryCrossingPrimeMeridian(geometry):
+    
+    primeMeridian = shapely.geometry.LineString([(0.,-90.),(0.,90.)])
+    featureShape = shapely.geometry.shape(geometry)
+
+    if(not featureShape.intersects(primeMeridian)):
+        return
+  
+    eastMask = shapely.geometry.Polygon([(0.,-90.),
+                                         (180.,-90.),
+                                         (180., 90.),
+                                         (0., 90.),
+                                         (0., -90.)])
+                                        
+    westMask = shapely.geometry.Polygon([(-180.,-90.),
+                                         (0.,-90.),
+                                         (0., 90.),
+                                         (-180., 90.),
+                                         (-180., -90.)])
+
+    print "This feature crosses the prime merdian."
+    print "  bounds before split:", featureShape.bounds
+
+    westShape = featureShape.difference(eastMask)
+    print "  bounds of western half:", westShape.bounds
+    eastShape = featureShape.difference(westMask)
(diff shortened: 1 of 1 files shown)


xylar:
An easy test of this script would be something like:
`./fix_features_at_prime_meridian.py -f landice/region/Antarctica_IMBIE5/region.geojson`
The result is stored in `features.geojson`.  To see a plot, do:
`./plot_features.py -f features.geojson -o test.png -m southpole`
There should be a darker line down the prime meridian, indicating that the feature has been split.

xylar:
@stephenprice, this new script should do what you need.  Test it out and merge the PR if you're happy with it.  #28 adds a driver script that merges all Antarctic basins into a single file and then uses this script to split all the basins that cross the prime meridian.

stephenprice:
@xylar, am I right in understanding that I can do this two ways (1) using this script to merge everything, then look for features that need to be split, and split them - or - (2) just manually split the features myself that I know need splitting? Thanks -

xylar:
@stephenprice, you have more or less the right idea but let me clarify how I would use this script in a typical workflow.  You should always do (1) above unless you're interested in a single basin.  In either case, you should never need to worry about whether the feature crosses the prime meridian or not.  The script detects that automatically so there is no harm in running all basins through `./fix_features_at_prime_meridian.py`, as I do in the example below.

My thinking is that we shouldn't alter the existing regions at all because it would be better to have a fix where MPAS supports crossing the prime meridian than splitting the features permanently.  Part of our goal here is to have features that are useful to the broader community and I think that goal is better supported if we don't alter the regions in the repository in ways that are specific to our project's needs.

Let's say you want all the east- and west-Antarctic basins (but not the peninsula) to cut them out of an MPAS grid.  Here would be the workflow I would use for that:
```bash
rm features.geojson
./merge_features.py -d landice -t WestAntarcticaIMBIE
./merge_features.py -d landice -t EastAntarcticaIMBIE
./fix_features_at_prime_meridian.py -f features.geojson
mv features.geojson EastAndWestAntarctica.geojson
./plot_features.py -f EastAndWestAntarctica.geojson -o EastAndWestAntarctica.png -m southpole
```
Then, you could use `EastAndWestAntarctica.geojson` as a mask for MPAS.

stephenprice:
@xylar - So that I'm clear, you are suggesting that the geojson files we've already got in the repo do NOT get changed in the repo, but that we only change them (e.g., alter the coords. in the file to allow for the split across the prime merid.) when we incorporate them in our own workflow (to define masks, etc.)? I ask for clarification because it looks like IMBIE basin 17 has already been split in the version that is now in the repo. Thanks -

xylar:
@stephenprice,  Basins in the repo needed to be split at the antimeridian for the general user.  This is because they don't display correctly in common tools unless they are split at -180/180 and because a lot of tools assume they are on a lat/lon map from -180 to 180 and -90 to 90.  So those changes have been made in the repo.

The split at the prime meridian is another story as far as I'm concerned.  MPAS made an arbitrary (and probably poor) choice to use 0 to 360 as the longitude range and this is causing trouble in the specific case of applying these regions as masks in MPAS.  But I don't think we want to alter the repo just because of this glitch.  Rather, @douglasjacobsen has suggested that we consider switching MPAS to use the more standard longitude range, and that we use a splitting tool like what I just made as an interim solution.

stephenprice:
@xylar  - my progress on this has been a bit stymied as I'm having to update my python to work w/ shapely. That, in turn, requires upgrading my macports, which is taking much longer than expected.

xylar:
@stephenprice, consider switching to anaconda, which has its own (much better for python) package manager.  @douglasjacobsen or @pwolfram or several others at LANL could undoubtedly help.

stephenprice:
@xylar - Ok, I follow on your comments above r.e. which features should get split in the repo vs. which should get split during the workflow. These nuances would likely never have occurred to me, so I'm glad someone's thought about it carefully. 

I can confirm that the 'fix_features_at_prime_meridian' script works as you suggest (my python is up to date for this now), which was the original issue w/ this PR, so I'll go ahead and merge it.

PR #28, opened 2016-02-25
Add a driver script for merging Antarctic basins

xylar:
The script combines all landice regions with names starting with Antarctica_IMBIE (because there is not a convenient tag for this) and splits any basins that cross the prime meridian into east and west parts.  Optionally (with --plot flag), the basins are plotted using the southpole map projection.

Also, renames setup_basins.py to setup_ocean_basins.py avoid confusion.

Additionally, this merge modifies the fix_features_at_prime_meridian.py so the western half of features are moved betwen 180 and 360 longitude, consistent with the MPAS convention that longitude goes from 0 to 360.


xylar:
@stephenprice, please merge #27 before merging this PR, as the driver script added here references the fix_features_at_prime_meridian.py script added by #27.

xylar:
To test, simply copy copy `fix_features_at_prime_meridian.py` into this directory (e.g. from master once #27 is merged).  Then run:
`./driver_scripts/setup_antarctic_basins.py --plot`
The result should be a file `Antarctic_Basins.geojson` that contains all the basins, split across the prime meridian, and a file `Antarctic_Basins.png` that is a plot of the basins.

stephenprice:
Thanks Xylar. I'll try to get to this later this evening or tomorrow.

stephenprice:
@xylar - in response to your "to test ..." comment above, I was able to generate the 'Antarctic_Basins.geojson' file, but w/ the '--plot' specification on there, I got a python error:

`Splitting features at prime meridian
  File "./driver_scripts/setup_antarctic_basins.py", line 42, in <module>
    subprocess.check_call(args, env=os.environ.copy())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 535, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory`

I'm assuming this could be some issue w/ my local installation. If I leave off "--plot" but then just use './plot_features.py -f Antarctic_Basins.geojson -o test.png -m southpole', I get the following image:
test

... which looks like yours, aside from the fact that there's only a bold line along the antimeridion (perhaps that is expected, but thought I'd check).

xylar:
@stephenprice, I'm guessing you didn't merge this branch with master, so you're missing the necessary `fix_features_at_prime_meridian.py` script (or maybe it doesn't have execute permission)?  It's definitely not working correctly because it didn't split the prime meridian.  That was the whole purpose of the exercise.

stephenprice:
@xylar - ok, let me try again ... hang on.

stephenprice:
@xylar - when I merged and re-tried, I got a few warnings, but no errors, and it was able to generate the plot correctly:

antarctic_basins

However, if I then try to use this features file to extract from one of our existing MPAS-LI Ant. meshes, I'm still seeing weird behavior (the culled regions don't look like I'd expect). 

Perhaps my workflow is flawed. To test, I did the following:

```
./merge_features.py -d landice -t EastAntarcticaIMBIE
./fix_features_at_prime_meridian.py -f features.geojson 
./plot_features.py -f features.geojson -o features.png -m southpole
```
The plot I get looks right - all the features that cross the prime and anti- merid. are split (as indicated by the bold lines). But when I use that same 'features.geojson' file to create a mask and cull from our mesh, here's what I see for the -m and -i culling options:

screen shot 2016-02-26 at 3 40 02 pm

I don't think this is expected, but copying @douglasjacobsen here in case he wants to weigh in. But it could be that for the purposes of this PR, the first plot above indicates that the tool is working as it should?

xylar:
@stephenprice, I don't really have a way to test the region stuff because I don't have an Antarctic mesh.  I would agree that things aren't working as they should, and I suspect it's still to do with the periodicity.

Probably the tool I made could move the fraction of each split feature that is currently between -180 and 0 lon and move it to be between 180 and 360 instead.  If the tool does that, it would be especially important that the features that are mangled in this way not get split and pushed back to the repository.

Let me give that a try.

xylar:
@stephenprice, I just added a small fix to the script you just merged in #27 that puts the western half of each feature between 180 and 360, so this should hopefully make MPAS happier.  Try out your test from before.  The plot came out looking the same for me, so I think the change did no harm at least...

By the way, I think the warnings and output you're getting are expected.  If you're concerned, post the output here and I can verify.  I'm getting a warning when I use cartopy that the artist subpackage of matplotlib is deprecated.  Nothing we can do anything about.

stephenprice:
@xylar - that looks like a fix to me. Here's some figs where I culled based on regions 2, 3, and 5, which are the only problematic ones w.r.t. the prime meridion ...

screen shot 2016-02-26 at 4 36 12 pm

And here's an example of culling out from the entire mesh (e.g., if I wanted to get rid of the buffer region we currently have in there:

screen shot 2016-02-26 at 4 42 01 pm

So, I think we're good with this one. Thanks! Jeez ... you should probably get to bed now.

PR #29, opened 2016-02-26
Fix typo Penninsula --> Peninsula in landice tag

xylar:
Fixes the tag "AntarticPeninsulaIMBIE", where peninsula was previously spelled with a double n.

### DIFF
--- a/landice/region/Antarctica_IMBIE24/region.geojson
+++ b/landice/region/Antarctica_IMBIE24/region.geojson
@@ -5,7 +5,7 @@
 		 "properties": {
 			"name": "Antarctica_IMBIE24",
 			"component": "landice",
-			"tags": "AntarcticPenninsulaIMBIE;WilkinsGeorgeVIceShelfDrainage",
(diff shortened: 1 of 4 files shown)


xylar:
@stephenprice, this PR might be a good chance to add some other tags to the Antarctic basins if there are others you can envision using.  Maybe you have it covered already with tags like "EastAntarcticaIMBIE" and "AmeryIceShelfDrainage".  Let me know if there would be other useful tags you could conceive of adding and I'll be happy to do it.

stephenprice:
@xylar , if you look at most of the basins, you should see that there area already a fair number of "standard" tags in there already (e.g., basin 21 has the following tags: "ThwaitesDrainage; PineIslandThwaitesDrainage; WestAntarcticaIMBIE"). I did a quick pass through on the obvious ones when I added them. I'm sure there are more, but I think what's there is probably sufficient for the time being. But if you see others that would also be handy to you, feel free to add them.

xylar:
@stephenprice, okay, sounds good.

PR #30, opened 2016-03-22
Add preservation of all properties in a feature

douglasjacobsen:
This ensures that all properties of a feature are preserved when merging
/ splitting them between feature files.


douglasjacobsen:
@xylar This should help in the process of using these to deepen critical passages.

xylar:
I checked that a new property ('depth' with value '100.0') added to the SouthernOcean region of the ocean was preserved during both splitting and merge operations.  Nice work, Doug!

xylar:
@douglasjacobsen, wince keys in the 'properties' dictionary are now alphabetized, it might make sense to split and merge all features once and commit them, so future split/merge operations don't appear to have made more changes than they actually have made.

PR #31, opened 2016-03-22
Alphabetize keys for all region properties

xylar:
This merge simply merges, then re-splits all features
in the database so that property keys are now alphabetized.
This way, future commits involving merge/split operations
are not complicated by this alphabetization.


xylar:
@douglasjacobsen, since nothing of substance is actually changed by this merge, it should be very easy to review, I hope.

douglasjacobsen:
@xylar Thanks for doing this.

PR #32, opened 2016-03-28
Add natural earth region for land coverage

douglasjacobsen:
This merge adds a region which defines where land exists, as defined by
the natural earth project (http://www.naturalearthdata.com/).


douglasjacobsen:
@xylar You don't have to merge this if you don't want to. This just adds a land region which can be used to cull an MPAS mesh based on where land is defined, rather than trying to do it with ETOPO.

xylar:
@douglasjacobsen, I'm happy to review this. 

The main issue I see just viewing the data in the geojson viewer is that this data set includes the Antarctic ice shelves in the ocean.  This is not the behavior we want when we're running without ice-shelf cavities and it also won't work quite right when we do have ice shelves because I doubt the grounding lines will match the data set that I'm using to produce them.  (I've made several modifications to the data, including smoothing it and sampling it to a coarser grid.)  I would suggest extracting the outline for the Antarctic coast from ETOPO1 and merging that with the land mask you have for every other region.

xylar:
@douglasjacobsen, having given this a bit more thought, I think the right solution is to take Antarctica entirely out of this map and to mask it out separately (either based on ETOPO1 or on Bedmap2, depending on if we have ice shelves or not).  I'll be happy to create the ETOPO1 and Bedmap2 regions so they can be added in a separate PR.

douglasjacobsen:
@xylar I think that's a good solution as well. I'm not in any hurry to get this PR satisfied, so if you want to make those two let me know and I can pull them into this PR.

xylar:
Merged.  @douglasjacobsen, you'll need to delete the remote, since I don't have permission.

PR #33, opened 2016-03-29
Add Antarctic land masks with and without ice shelves

xylar:
Adds land masks extracted from the Bedmap2 data set that are either based on the ice mask or the grounded ice mask. The masks should include all land south of 60S latitude.


xylar:
@douglasjacobsen, these two regions would be more likely to serve our needs for the Antarctic land masks either with or without ice-shelf cavities.  I'm not exactly sure how we want to combine these masks with those from #32.  It might make sense to leave #32 as it is and use a mask that covers the region south of 60S to exclude that region before applying one of these masks.  Let me know what you think.

xylar:
Here is the script I used to extract the Bedmap2 data that had been interpolated to the 1/60 degree ETOPO1 grid:
https://gist.github.com/xylar/d8095949cd05d74e613c

xylar:
I'm going to close this PR and delete this branch because I think it makes more sense to combine this with #32.

PR #34, opened 2016-04-04
Add simplify_features script

xylar:
Uses shapely's simplify command to remove redundancies in features.
The tolerance flag can be used to allow more simplification by
allowing features to move within a tolerance.  This may be appropriate,
for example, if the shape was adapted from a contour of a dataset on
a grid (e.g. the Bedmap2 land coverage masks).

### DIFF
--- a/simplify_features.py
+++ b/simplify_features.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+"""
+This script takes a file containing one or more feature definitions, that is
+pointed to by the -f flag.  Features in the input file are simplified
+using the shapely library with a tolerance (in degrees lon/lat) given
+by -t flag.  If the -t flag is omitted, the default tolerance is zero
+(so that points will not be allowed ot move at all, but may still be
+simplified).  The result are stored in features.geojson.
+"""
+
+import json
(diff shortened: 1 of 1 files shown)


xylar:
@douglasjacobsen, no rush on this one. I needed this script for cleaning up the Bedmap2 data, and figured it would be useful in other cases, too.

simplify_features.py
```diff
@@ -0,0 +1,84 @@
...
+import shapely.geometry
+import shapely.ops
+
+parser = argparse.ArgumentParser(description=__doc__,
+                                 formatter_class=argparse.RawTextHelpFormatter)
+parser.add_argument("-f", "--feature_file", dest="feature_file",
+                    help="Feature file to be simplified", metavar="FILE",
+                    required=True)
+parser.add_argument("-t", "--tolerance", dest="tolerance", type=float, default=0.0,
+                    help="A  feature whose overlap with the first feature should be removed",
```
douglasjacobsen: This doesn't seem like the right help message to me.

xylar:
@douglasjacobsen, I fixed the bad help message.  Good catch!

PR #35, opened 2016-04-04
Simplify the Bedmap2 land coverage regions

xylar:
Simplification was performed with the new simplify_features.py
script with a tolerance of 0.02 degrees.


xylar:
The files are a lot smaller, should process faster, and look cleaner without losing information beyond the scale of the grid that was used to create the original contours.

PR #37, opened 2016-04-05
Adds link to provenance for ocean regions

pwolfram:
We didn't cite our data source from http://www.marineregions.org/images/boundaries/IHO_crop.png.  This fixes this oversight. 

cc @akturner

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,12 @@
-This repository will house definitions of geometric features. These features
+This repository houses definitions of geometric features. These features
 can include regions, transects, and points. Currently they are described in
-geojson format.
+geojson format.  For example, here are some regions for Antarctica.
+
+![alt text](https://cloud.githubusercontent.com/assets/4179064/12921663/93282b64-cf4e-11e5-9260-a78dfadc4459.png "Antarctica regions")
+
+Data for regions has been derived from sources listed in the
+[contributors file](contributors/CONTRIBUTORS.md) as specified
+via the `author` field in each `geojson` file.
 
 The scripts in the top level directory can be used to help maintain and use
 this repository. All of them can be run with the -h flag to get usage
--- a/contributors/CONTRIBUTORS.md
+++ b/contributors/CONTRIBUTORS.md
@@ -1,0 +1,13 @@
+# Data sources
+
+Data sources and authors of the regions may be updated from the repository via
+this [script](list_contributors.sh):
+
+List populated on 2016-04-05 13:10:
+
+ * Todd Ringler
+ * Xylar Asay-Davis
+ * http://icesat4.gsfc.nasa.gov/cryo_data/ant_grn_drainage_systems.php
+ * http://www.marineregions.org/downloads.php#iho
+ * http://www.naturalearthdata.com/
+ * https://www.bas.ac.uk/project/bedmap-2/
--- a/contributors/CONTRIBUTORS_HEADER
+++ b/contributors/CONTRIBUTORS_HEADER
@@ -1,0 +1,5 @@
+# Data sources
+
+Data sources and authors of the regions may be updated from the repository via
+this [script](list_contributors.sh):
+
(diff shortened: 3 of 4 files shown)


pwolfram:
@xylar and @akturner: it may also be good to add to the README.mx file provenance for land/sea ice regions used too.

douglasjacobsen:
@pwolfram I actually don't think we want this PR. The provenance information is contained within the geojson files (see: https://raw.githubusercontent.com/MPAS-Dev/geometric_features/master/ocean/region/Adriatic_Sea/region.geojson). Since each region definition can come from a different source.

I don't think we want to list all of the sources we use in the README file, but I could be wrong about that. Either way, if that's what we want, we should parse all of the region files we have, and put the provenance information into the README file rather than this single citation.

douglasjacobsen:
Also, not all of the ocean stuff came from that single source. See: https://raw.githubusercontent.com/MPAS-Dev/geometric_features/master/ocean/region/Southern_Ocean/region.geojson which @xylar created.

pwolfram:
@douglasjacobsen, I think we need to have some type of information in the README that specifies where the data source originated.  As it is now, it could be perceived as bland and unappealing for a new user and it is unclear where the information originated.  I like your idea to

> I don't think we want to list all of the sources we use in the README file, but I could be wrong about that. Either way, if that's what we want, we should parse all of the region files we have, and put the provenance information into the README file rather than this single citation.

but I also think we should have a visual on the page to help visually illustrate its capabilities.

As long as we: 
 1. Provenance the origins of the data in a short form way
 2. Provide a visual to demonstrate capabilities

the intent of this PR will be fulfilled.  I'm less interested in the specifics used to meet these goals.  However, I really do think we need to meet these objectives for clarity.

douglasjacobsen:
I think a more appropriate place for listing the origins of the data would be in a CONTRIBUTORS.md file. I don't think it's as important to have a high level list of where the data came from, since that will be difficult to maintain in a useful way, but if other people feel like it's important and you are all willing to update this PR to include all of the provenance information, I'm open to reviewing it after that's complete.

I'm not sure how to provide a visual to demonstrate the capabilities of this repo, but if you have an example of what you think would be useful, I'd be open to adding it to this.

pwolfram:
Done, check out the latest...  to update the `CONTRIBUTORS.md` file simply run 

`./contributors/list_contributors.sh`.

pwolfram:
FYI @xylar, I'm borrowing one of your images for the "logo".

pwolfram:
You can easily see how this looks at https://github.com/pwolfram/geometric_features/tree/data_provenance

douglasjacobsen:
Thanks for these changes.

Two comments. First, can you remove the addition of the citation in README.md. And second, can you change the bash script to python to be more consistent with the other scripts we've been working on?

douglasjacobsen:
@pwolfram Sorry, I misread the README.md change. I think that change looks good, so ignore that part of my last comment.

pwolfram:
@douglasjacobsen, I'm not sure the benefit to using python here when this is a task is readily done via the shell.  I can always wrap the shell command in python if you really want but this seems silly.

pwolfram:
@douglasjacobsen, after thinking about this more I agree because this helps standardize and makes it easier to use later.

pwolfram:
@douglasjacobsen, feel free to check it out now.

douglasjacobsen:
@pwolfram Looks good. Thanks!

xylar:
@pwolfram, thanks for doing this.  I think the image is a good choice and I agree that it's useful to have the sources listed.

I noticed that the link in CONTRIBUTORS.md is still to the .sh script, not to the new .py script.  That should be fixed.

pwolfram:
Thanks for catching this @xylar, the fix is at https://github.com/MPAS-Dev/geometric_features/pull/38.

pwolfram:
Thanks @douglasjacobsen and @xylar!

PR #40, opened 2016-06-25
Add a script that subdivides transects

xylar:
This PR adds a script, subdivide_transects.py, that is used
to subdivide transects (LineStrings) longer than a maximum length
(in degrees) into an integer number of subsegments. This is useful
because some tools (e.g. the MPAS MaskCreator) don't necessarily
follow the desired path between points that are far apart.

The maximum allowed segment length is supplied (in degrees) with
the -m flag at the command line.

### DIFF
--- a/subdivide_transects.py
+++ b/subdivide_transects.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+
+"""
+This script takes a file containing one or more feature definitions,
+pointed to by the -f flag and a flag -m indicating the maximum length of
+segments of transects in degrees.  Segments of transects (LineStrings)
+that are longer than the maximum are subdivided into an integer number
+of subsegments so the minimum is reached. The resulting features are placed
+in (or appended to) features.geojson.
+"""
+
+import json
+import argparse
+from collections import defaultdict
+from utils.feature_write_utils import write_all_features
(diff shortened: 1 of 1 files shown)


xylar:
@douglasjacobsen, this PR isn't a high priority but it also shouldn't be hard to test.  I found that the MaskCreator makes a weird zig-zag between points that are spaced far apart (e.g. going 3 grid cells in one direction, 3 in another, etc.) rather than trying to follow as closely as possible the great circle path between points.  Subdividing LineStrings seems to give me behavior that's at least close to what I want (though following straight lines in lat/lon space rather than great circles).

pwolfram:
@xylar, would you like me to review this PR?  I'm thinking it should be merged before #52, which we are using partially as a clean-up type step.  Would you concur?

But, based on the above comment it sounds like there may still be an issue with this PR, is that true?

xylar:
@pwolfram, I'm not yet sure this is useful to anyone but me, so let's leave it alone for now.

I'm not sure anyone else is trying to use the mask creator with transects, and I have already subdivided the transects used to open critical passages so they don't have this issue.

xylar:
An updated version of the script has been added to #78.

Issue #41, opened 2016-09-09
script to join a number of regions together to make one single region


milenaveneziani:
@xylar, @pwolfram, @douglasjacobsen: tugging you on this, as you are probably most interested to provide input here.

Yesterday, I was trying to create a mask file for ocean basins, and specifically for 2 groups: oceanBasins and oceanBasinsMOC. These groups contain the ocean basins that can now be reproduced by using the drivers that @xylar wrote a few months back. 
Let us take Atlantic_MOC as an example. 
The problem is that the corresponding geojson file for this region is a collection of the subregions that make up Atlantic_MOC, whereas what we really need is a single region, called Atlantic_MOC, that is the result of the joint subregional polygons.
This way, say for the oceanBasinsMOC group, we will have 2 single regions: Atlantic_MOC and IndoPacific_MOC.
Does this make sense?

Would it be hard to write such a script?

xylar:
@milenaveneziani, I get what you're saying.  Presumably rather than going through the trouble of joining the geometry into a single region, what is needed is a way to collect geometry from multiple regions together and give it a different name.  I can't recall if something to do that already exists but I think so. @douglasjacobsen, can you comment?

@milenaveneziani, can you be more specific about what the needs are?  How exactly are you using the geojson files so I understand whether it matters that the geometry itself gets fused together as opposed to just giving a single name to the collection of geometry.

milenaveneziani:
@xylar: I talked with Doug yesterday (he explained the whole process to me, which is why I came to the conclusion that we needed this issue open), and he doesn't think there is such a script now.
We need one single, fused region because we will need to compute specific quantities on each of them. For example MOC or meridional heat transport: we will want to compute these quantities over Atlantic_MOC and IndoPacific_MOC as whole regions, not as a collection of subregions. Same goes for the simple oceanBasins groups: we will want a maskfile for the fused Atlantic, SO, etc.

@douglasjacobsen thought that we shouldn't create these regions once and for all, but that it would be better to have a driver (similar to the one you have to merge the features) that would join the regions. There should of course be a provenance of how say Atlantic_MOC was created at a given time.

xylar:
@milenaveneziani, okay, I think I see.  I don't think you mean "fused" in the same way I mean "fused".  It doesn't sound like it matters whether the combined region is made up of multiple polygons or one single polygon, for example.

This should be quite easy to do.  I'll see if I can bang it out this evening.

milenaveneziani:
[...]

PR #42, opened 2016-09-09
Add a script that combine features into single feature definition

xylar:
The script takes a file containing one or more feature definitions,
and a new name for the combined feature. The geometry of the
feature definitions are combined into a single feature definition,
which is placed in (or appended to) features.geojson.

The script for creating MOC basins has been updated to use
this script combines all the sub-basins into single feature
definitions named "<basin>_MOC".


xylar:
# Testing

I ran `./driver_scripts/setup_MOC_basins.py --plot` but the plotting has some artifacts (as mentioned in #41).

Here are the resulting geojson files:
https://gist.github.com/xylar/5cda1cf32c80c0ab5259dda1a0ab1de5

Seems to work exactly as desired: It produces a single geometric feature with a single name.

xylar:
I also cleaned up the `setup_MOC_basins.py` script to be less redundant.  It uses dictionaries to define which tags ("sub-basins") are in each MOC basin and which mask is used to cut off the southern limit of the basin.

xylar:
@douglasjacobsen and @pwolfram, you might also want to take a look at this.  (Especially with regards to coding style, docstring, etc.)

xylar:
@milenaveneziani, are there other driver scripts that also need to be updated to use this new script?

xylar:
This PR is intended to address #41

pwolfram:
@xylar, I have some ideas for improvement on at least style and documentation if you are still interested in them (assuming this doesn't have to be merged ASAP).  If this is helpful, do you want me to outline them for you or just make the changes via a PR to your branch and then you can take a look?

xylar:
@pwolfram, just make a commit directly to this branch with your suggested changes.  I'm headed to bed so I can take a look at what you've done in the morning (and tweak if need be).

pwolfram:
@xylar, here are some changes that help clarify the script and simplify it.  In particular, we may need to do even further refining than in `e9c94ff`.  I tested to make sure these results were bit-for-bit relative to your previous branch.

milenaveneziani:
@xylar and @pwolfram: thanks for doing this.
Another, similar driver is setup_ocean_basins, which defines 6 basins: Atlantic, Pacific, SO, Indian, Mediterranean, Arctic. I believe all of them are from merging different features, which should also be joined/fused.

BTW: I have some second thoughts about the northern boundary of the Atlantic_MOC region (IndoPacific_MOC is fine). I have pinged Mat Maltrud to ask what he used in POP for MOC and MHT calculations: I will report back on Monday about this.
Tagging @mark-petersen as well, as he may have some feedback about the way he computed Atlantic MOC in MPAS before.

xylar:
@pwolfram, thanks for the cleanup work.  I made sure the output option and author/modified docstrings are in all the relevant manipulation scripts.  I think the output argument is really useful and does make the driver scripts a lot cleaner.

xylar:
@milenaveneziani, to address the "provenance" request in #41, I have added a "constituents" property in the combined feature that is a list of the region names that were used to produce it.  Let me know if that addresses that request.

I updated the driver for producing ocean basins as well.  The updated results from both driver scripts can be seen here:
https://gist.github.com/xylar/28efbd4f1a401cc8f95bfb8cbe7d1607

As far as I'm concerned, this PR is ready to be merged whenever you are ready.  Let me know if you have any other requests or concerns.

I believe we're being a bit looser with the merge requirements for this repository than for other MPAS repos.  That means you can merge just by hitting the "merge pull request" button, rather than doing it with the command line.  @pwolfram and @douglasjacobsen, correct me if I'm wrong.

douglasjacobsen:
@xylar Yes. We are able to use the green merge button in this Repo.

Thanks for taking the time to write this script. It (seems) to be exactly what @milenaveneziani was looking for.

xylar:
Great, thanks, @douglasjacobsen!

milenaveneziani:
ok, @xylar. I looked closely at all basin regions and also got feedback from @maltrud. Here are my questions:
1) I don't know if it's a plotting issue again, but several islands are still being considered as ocean in all domains. The question is: does this really matter? I guess it depends on how mask_creator operates. In any case, this may be a more delicate issues for the MOC and MHT calculations.
2) the northern boundary of Atlantic_MOC is fine. But I now see that the southern boundary is shifted to the north with respect to what it was before this merging. Maybe one slice of South Atlantic is missing? The southern boundary should go to the tip of South Africa (30S, I believe).

xylar:
@milenaveneziani,

> 1) I don't know if it's a plotting issue again, but several islands are still being considered as ocean in all domains. The question is: does this really matter? I guess it depends on how mask_creator operates. In any case, this may be a more delicate issues for the MOC and MHT calculations.

Yes as I said above, it's a plotting artifact and not a problem with the geojson files themselves, as you can see by going to the gist where I have all the resulting geojson files.  There, the plotting looks correct.

> 2) the northern boundary of Atlantic_MOC is fine. But I now see that the southern boundary is shifted to the north with respect to what it was before this merging. Maybe one slice of South Atlantic is missing? The southern boundary should go to the tip of South Africa (30S, I believe).

 I have to say I'm not sure what you mean.  If you look both at the code and at the results here:
https://gist.github.com/xylar/28efbd4f1a401cc8f95bfb8cbe7d1607#file-atlantic_moc-geojson
it looks clear to me that the Atlantic_MOC region is being cut off at 30S as before.  If you want a different latitude, that can be arranged but I think the results are the same as before.

milenaveneziani:
@xylar:
- about 1) above, I am looking at the gist link (this, to be sure: https://gist.github.com/xylar/28efbd4f1a401cc8f95bfb8cbe7d1607): this should have the plotting correct, right? But if you zoom in, you will still see islands that are considered ocean in all basins (it's easier to see it in the Arctic).
- above 2): never mind, I re-checked your earlier PR's and looked at whole the maps, and the reasons we decided to go with 30S was to make things consistent between Atlantic_MOC and IndoPacific_MOC. We could choose a southern boundary of ~32S for the Atlantic, but that wouldn't work for the IndoPacific_MOC and Pacific_MOC basins. Sorry about the confusion.

maltrud:
@milenaveneziani : why can't we change both Atl and Indo-Pac MOC masks to extend to 32S?

milenaveneziani:
@maltrud, @xylar: I am sorry, I am going back and forth on this. I thought that by doing that, we would cut through the Great Australian Bight (and we do), but that's OK because that is Indian Ocean anyway..

douglasjacobsen:
@maltrud and @milenaveneziani: It'd be fine to change the region definitions, but that should be a separate discussion. This PR is only about adding the new script capability, not necessarily making sure the region definitions are what you guys want.

@milenaveneziani As far as this PR goes, do you have any other concerns about it? Specifically about the provenance information?

pwolfram:
Note, #43 considers this issue of islands being plotted as ocean

milenaveneziani:
@douglasjacobsen, @xylar: yes, provenance is great.
And I agree that there should be a different PR for the region definition.
So, the only thing left is, *if* the maps on the gist link are correct, checking why some islands are still covered by ocean.

douglasjacobsen:
@milenaveneziani: My guess would be that those islands are covered in the original region definitions. Do you have an example of an island that is newly covered after this PR, but was not covered prior to this?

We could take a look at that if it is, but my guess is that none of the coverage is new and then if we want them to not be covered, that should be also part of another PR.

milenaveneziani:
@douglasjacobsen: you are right, the islands are also not there in the original subregions.
Maybe there is an exception though, and that is the Southern Ocean region: the merged product doesn't seem to consider Tasmania and New Zealand as islands, but the original regions seem to have those:
https://cloud.githubusercontent.com/assets/4179064/12974706/4623f96a-d0b6-11e5-9988-f9e03b315d39.png
Do you understand why?

milenaveneziani:
No, I don't think we noticed the island things in #20.
Let me just check with @mark-petersen to see if they matter in the MOC and MHT integral calculation.

douglasjacobsen:
@milenaveneziani They won't matter, because the ocean grid has land removed. So in the resulting ocean masks file, there would only be ocean cells anyway.

milenaveneziani:
ok, sounds good.

xylar:
Thanks, @milenaveneziani!

Issue #43, opened 2016-09-12
Plotting script does not exclude islands


pwolfram:
As discussed in https://github.com/MPAS-Dev/geometric_features/issues/41, our current plotting script does not exclude holes in Polygons and consequently shades islands.

cc @xylar

xylar:
I'll look into this when I get a chance.  @pwolfram, if you have time to investigate, that would be good, too.    Polygons and multipolygons are stored as a list of shapes, the first of which is the outline and the rest of which are the holes.  I suspect that only the first polygon is being used or that the holes are being treated incorrectly as normal polygons.  But I haven't investigated.

It would be nice if we didn't have to do so much of the plot setup ourselves.  For example, there's a library called Descartes that can do standard matplotlib plots directly from shapely objects but I don't think it can do fancy transformations and maps like we're using.

pwolfram:
@xylar, you are right that the first list of points in `coordinates` is the outer polygon and the next items in the list are holes.

I started to investigate this over the weekend but didn't fully get to a viable solution.  I would like to have one more stab before sharing my code.  How quickly do we need this fix @xylar and @milenaveneziani?

It appears we need to use something like `descartes` to convert the polygons to `matplotlib` patches but I wasn't able to get this to work properly and probably need more time for debugging.

pwolfram:
Preliminary work is at https://github.com/pwolfram/geometric_features/tree/fix_island_plotting_bug

douglasjacobsen:
@pwolfram This should be a low priority, as there are alternate methods of plotting that capture the holes correctly. But I think you can spend whatever time you want on fixing it.

xylar:
Yep, super low priority.  It's just a convenience that we'd like to be able to plot our regions to verify that we're producing what we want.  Given we can use http://geojson.io/ or GitHub itself to view the files, it's not critical.

xylar:
@pwolfram, one other plotting issue you might look into is that I accidentally merged some transects and regions into the same geojson file (happened to be everything with the tag `Mediterranean_Basin`) and it looked like the transects were getting plotted as polygons instead of lines.  Presumably this is a separate bug but while you're working on plotting...

If you can't recreate this issue, let me know and I'll try.

pwolfram:
@xylar, can you please double-check to see if this transects issue still exists following #45?  I think this should fix this issue too...

xylar:
This issue is addressed by #48.

PR #45, opened 2016-09-12
Fixes issue with islands being plotted as water

pwolfram:
Addresses #43 to use descartes / shapely to convert geojson files into matplotlib patches that support holes, e.g., islands.

cc @milenaveneziani and @douglasjacobsen

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -99,14 +99,13 @@ def divide_poly_segments(points): #{{{
 
     return outPoints #}}}
 
-def plot_poly(mapInfo, points, color): #{{{
+def plot_poly(mapInfo, geom, color, filled=True): #{{{
 
     refProjection = cartopy.crs.PlateCarree()
 
-    if type(points) is dict:
-        simplepoints = shapely.geometry.shape(points)
-    else:
-        points = divide_poly_segments(points)
+    polygon = not (type(geom) == np.ndarray)
+    if not polygon:
+        points = divide_poly_segments(geom)
         x = points[:,0]
         y = points[:,1]
 
@@ -114,10 +113,17 @@ def plot_poly(mapInfo, points, color): #{{{
 
     for mapType in mapInfo:
         (ax, projection, plotFileName, fig) = mapInfo[mapType]
-        if type(points) is dict:
-            ax.add_patch(PolygonPatch(simplepoints, **plotprop))
+        if polygon:
+            if type(geom) == shapely.geometry.multipolygon.MultiPolygon:
+                for apoly in geom:
+                    ax.add_patch(PolygonPatch(apoly, **plotprop))
+            else:
+                ax.add_patch(PolygonPatch(geom, **plotprop))
         else:
-            ax.fill(x, y, **plotprop)
+            if filled:
+                ax.fill(x, y, **plotprop)
+            else:
+                ax.plot(x, y, '-', **plotprop)
 
     return #}}}
 
@@ -148,6 +154,7 @@ def plot_features_file(featurefile, mapInfo): #{{{
     for feature in featuredat['features']:
         polytype = feature['geometry']['type']
         coords = feature['geometry']['coordinates']
+        shape = shapely.geometry.shape(feature['geometry'])
         featurename = feature['properties']['name']
         print '  feature: %s'%featurename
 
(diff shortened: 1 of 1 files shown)


pwolfram:
Note, generalization to support MultiPolygons is needed and is in progress but this should work for the use case in #41.

plot_features.py
```diff
@@ -138,34 +148,41 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
-                    plot_poly(mapInfo,poly,colors[color_num])
+                try:
+                    patch = feature['geometry']
+                    plot_poly(mapInfo, patch, colors[color_num])
+                except:
+                    for poly in coords:
+                        plot_poly(mapInfo, poly, colors[color_num])
             elif polytype == 'LineString':
-                plot_poly(mapInfo,coords,colors[color_num],filled=False)
+                plot_poly(mapInfo, coords, colors[color_num], filled=False)
```
xylar: I realize this is really outside the scope of this PR but could you also make LineStrings get plotted as lines instead of polygons?  I believe even plotting them as unfilled polygons is wrong since they are sometimes (often?) transects, which are not closed shapes like polygons.

xylar:
@pwolfram, this looks promising.  I'll take a closer look tomorrow.  Does descartes support MultiPolygons and you just haven't got to that yet?  Or do we need to iterate through the polygons ourselves?  Either way, seems simple enough to add, right?  Might be worth waiting on this PR until that's addressed.

plot_features.py
douglasjacobsen: I agree. Keep in mind, `MultiLineString` should be handled in the same way. It should be an array of `LineString` type constructs, and none of them are required to connect or to close.

pwolfram:
@xylar, agreed on MultiPolygons.  I've got a fix in the works I'm testing now.

pwolfram:
@xylar, I've pushed the commit that supports this and you can readily test with the `IndoPacific_Basin.geojson`, e.g., 
screenshot 2016-09-12 15 59 19

Note, you can see that transect issue you were discussing earlier.  The issue is that this is a `MultiPolygon` so each of these is plotted as a polygon.  If a transect got in here it this should have been a multiple feature geojson file instead. 

For clarity, I'm using this file to test: https://gist.github.com/a10c8b1ff5a302940d44ac2d5dcb17b1

plot_features.py
pwolfram: We won't plot lines as filled polygons anymore and I'll fix this while I'm doing these edits.

pwolfram: This should be fixed now.

pwolfram:
@xylar and @douglasjacobsen, this should be ready to review.

Note, just to be 100% clear here-- this implementation is somewhat of a hack because it is not general.  What we really should be doing is pure plotting from a shapely definition of geometry (similar to what descartes does) and then just applying a plot method to it with modifiers on how it should be plotted.  However, this PR should work in the interim but I did open #46 in case we need to return to this in the future.

xylar:
@pwolfram, I agree that this could theoretically be made more general but I think your fix takes care of all the use cases I can envision.  If you have time to work on the more general version down the road, that would be great.  I don't know enough about descartes to know how hard that would be.

I haven't had a chance to really test this today but I'll get to it as soon as I can.

pwolfram:
@xylar, descartes is essentially a thin-wrapper that converts shapely and geojson like datastructures to matplotlib polygon patches.  It is ~50 lines of code (https://github.com/pwolfram/descartes/blob/master/descartes/patch.py) and not very general past polygons.  I could envision generalizing this so that an arbitrary geojson or shapely file could be readily visualized.  This is probably out-of-scope for now but may be something we need for convenience / efficiency in the future.  

An additional complication to be considered is that we need to supply transforms to handle geodesic data, which is not needed on the plane obviously.

milenaveneziani:
@pwolfram: here are the new Atlantic_MOC and IndoPacific_MOC with the plot_features in this PR. Seems like it works to me.

atlantic_moc
indopacific_moc

plot_features.py
```diff
@@ -132,40 +142,45 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
-            print 'Error plotting %s'%(feature)
+        except Exception as inst:
+            print 'Error plotting %s with error %s'%(featurename, inst)
 
         feature_num = feature_num + 1
 
     for mapType in mapInfo:
         (ax, projection, plotFileName, fig) = mapInfo[mapType]
         print 'saving ' + plotFileName
+        ax.set_global()
```
xylar: @pwolfram, am I right in thinking that this line has made it so the plots don't "zoom" in on the region being plotted but rather plot the entire globe every time?

I'm okay with that change but I want to make sure others are as well.  i could imagine cases where a given plot is quite regional and plotting the entire globe is overkill (e.g. when I've been focused on the Mediterranean) so maybe we want a command-line argument that lets the user force a plot to be "regional" if they want that?

plot_features.py
```diff
@@ -97,29 +101,35 @@ def divide_poly_segments(points): #{{{
...
-
-    return #}}}
-
-def plot_point(mapInfo, points, marker, color): #{{{
+    genprop = {'transform':refProjection, 'alpha':0.4, 'linewidth':2.0}
 
-    points = np.asarray(points)
+    # merge plot properties
+    if kwargs is not None:
+        plotprop = genprop.copy()
```
xylar: I think if `kwargs` is `None`, `plotprop` never gets assigned.  Since you're not reusing genprop for anything, and you just rename that to `plotprop` and not do the copy operation here?

xylar:
@pwolfram, in the process of looking at your modifications and trying to figure out how to make the map extent fit the geometry rather than being global, I came across:

http://scitools.org.uk/cartopy/docs/v0.13/matplotlib/geoaxes.html#cartopy.mpl.geoaxes.GeoAxes.add_geometries

It seems to me this is exactly what we need (instead of descartes) to support general shapely objects.  I have a much simplified version of the plotting script that makes use of this.  I will likely make a PR to this PR with those changes as soon as I figure out how to reduce the extent of my plots to the region of interest rather than global (which somehow seems to be the default).

pwolfram:
I'm closing this PR in favor of the simplified and generalized https://github.com/MPAS-Dev/geometric_features/pull/48 that uses `add_geometries`.  Great work on finding this @xylar!

PR #47, opened 2016-09-13
Modified Atlantic_MOC and IndoPacific_MOC features to extend to 34S.

milenaveneziani:
This commit modifies the mask for creating Atlantic_MOC and IndoPacific_MOC, so that they can extend to 34S.


milenaveneziani:
@xylar, @douglasjacobsen: here is the PR to modify the Atlantic_MOC and IndoPacific_MOC to extend to 34S. Let me know if this is ok.
I double checked with @maltrud and @mark-petersen and they said they should be OK for MOC/MHT calculations.

here is the png I created from plot_features (disregard the island issue):
atlantic_moc
indopacific_moc

pwolfram:
@milenaveneziani, this PR https://github.com/MPAS-Dev/geometric_features/pull/45 should fix the island issue-- do you mind testing plotting using this application above?

milenaveneziani:
Thanks @pwolfram. I tested the new plot_features and it works. I have attached the new figures to #45.

xylar:
@milenaveneziani, it looks like you didn't add `ocean/region/MOC_mask_34S/region.geojson` to the repository.  Otherwise things look good to me.  Thanks for adding the `.gitignore`.

xylar:
@milenaveneziani, it might also be good for you to make your own fork of the `geometric_features` repository rather making branches within `MPAS-Dev/geometric_features`.  @douglasjacobsen, maybe you could help with that process if you agree that's the preferred way?

For this PR, I definitely don't think it's a big deal.

milenaveneziani:
@xylar: I noticed that too (that the MOC_mask_34S/region/geojson files was not added). I am not sure why.. I can certainly see it in my local branch.
Is it possible that .gitignore looks in any subdirectory as well?

milenaveneziani:
Maybe that was the case.. I modified .gitignore to just ignore the png files. For some reason, that didn't work (the png files were added anyway).
Anyways, now things should be fixed and I can see the region.geojson file in MOC_mask_34S.

xylar:
@milenaveneziani, now it looks like you put back the original `MOC_mask_30S` and the `MOc_mask_34S` still isn't there.

It looks like there's a way to do what you want.  You can change the `.gitignore` to be:
```
\*.png
\*.geojson
```
That should only apply the pattern to the current directory.

milenaveneziani:
@xylar: sorry, I included the changes to the script now, and also the modification to .gitignore.

xylar:
@milenaveneziani, sorry, I somehow used `\` instead of `/` in my suggestion above.  I'm going to fix this myself.

Issue #46, opened 2016-09-12
Generalize plotting to work from geojson / shapely instance


pwolfram:
Plotting currently operates on geojson in an ad-hoc fashion.  To generalize we could:

 1. produce shapely instance from geojson file (readily done via something like 
`shape = shapely.geometry.shape(feature['geometry'])`)
 2. plot based on type of shapely instance (e.g., generalize existing [descartes library ](https://bitbucket.org/sgillies/descartes))

pwolfram:
I believe this should be handled in general by #48 (potentially following a minor edit to `plot_features.py`), would you concur @xylar?  If so, once we merge #48, we can close this issue.

xylar:
I think this issue is addressed sufficiently by #48.  Types other than `Polygon`, `MultiPolygon`, `LineString`, `MultiLineString` and `Point` are not supported elsewhere in this repo (e.g. in utils) so I don't think they should reasonably be expected to be supported by the plotting script.

xylar:
Oops, shouldn't close this until #48 is merged...

PR #48, opened 2016-09-14
Change plotting to use cartopy add_geometries

xylar:
This merge used cartopy support for shapely geometries to perform
plotting rather than creating plotting objects ourselves.

Also, plotting properties have been improved so that transects
are not plotted as filled shapes.

The bounds of the plot are computed based on the combined bounding
box of all plotted objects (other than the coastlines).

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -68,8 +68,58 @@ def plot_base(mapType, projection): #{{{
 
     return (ax,projection) #}}}
 
+def subdivide_geom(geometry, geomtype, max_length): #{{{
 
-def plot_features_file(featurefile, mapInfo): #{{{
+    def subdivide_line_string(lineString, periodic=False): #{{{
+        coords = list(lineString.coords)
+        if periodic:
+            # add periodic last entry
+            coords.append(coords[0])
+
+        outCoords = [coords[0]]
+        for iVert in range(len(coords)-1):
+            segment = shapely.geometry.LineString([coords[iVert],coords[iVert+1]])
+            if(segment.length < max_length):
+                outCoords.append(coords[iVert+1])
+            else:
+                # we need to subdivide this segment
+                subsegment_count = int(np.ceil(segment.length/max_length))
+                for index in range(subsegment_count):
+                    point = segment.interpolate(float(index+1)/float(subsegment_count), normalized=True)
+                    outCoords.append(point.coords[0])
+
+        if periodic:
+            # remove the last entry
+            outCoords.pop()
+        return outCoords  #}}}
+
+    if geomtype == 'LineString':
+        newGeometry = shapely.geometry.LineString(subdivide_line_string(geometry))
+    elif geomtype == 'MultiLineString':
+        outStrings = [subdivide_line_string(inLineString) for inLineString in geometry]
+        newGeometry = shapely.geometry.MultiLineString(outStrings)
+    elif geomtype == 'Polygon':
+        exterior = subdivide_line_string(geometry.exterior, periodic=True)
+        interiors = [subdivide_line_string(inLineString, periodic=True) for inLineString in geometry.interiors]
+        newGeometry = shapely.geometry.Polygon(exterior, interiors)
+    elif geomtype == 'MultiPolygon':
+        polygons = []
+        for polygon in geometry:
+            exterior = subdivide_line_string(polygon.exterior, periodic=True)
+            interiors = [subdivide_line_string(inLineString, periodic=True) for inLineString in polygon.interiors]
+            polygons.append((exterior, interiors))
+
+        newGeometry = shapely.geometry.MultiPolygon(polygons)
+    elif geomtype == 'Point':
+        newGeometry = geometry
+    else:
+        print "Warning: subdividing geometry type %s is not supported."%geomtype
+        newGeometry = geometry
+
+
+    return newGeometry #}}}
+
+def plot_features_file(featurefile, mapInfo, max_length): #{{{
 
     # open up the database
     with open(featurefile) as f:
@@ -86,6 +136,9 @@ def plot_features_file(featurefile, mapInfo): #{{{
     for feature in featuredat['features']:
         geomtype = feature['geometry']['type']
         shape = shapely.geometry.shape(feature['geometry'])
+        if(max_length > 0.0):
+            shape = subdivide_geom(shape, geomtype, max_length)
+
         featurename = feature['properties']['name']
         print '  feature: %s'%featurename
 
@@ -117,6 +170,8 @@ def plot_features_file(featurefile, mapInfo): #{{{
         feature_num = feature_num + 1
 
     box = shapely.geometry.box(*bounds)
+    if(max_length > 0.0):
+        box = subdivide_geom(box, 'Polygon', max_length)
 
     for mapType in mapInfo:
         (ax, projection, plotFileName, fig) = mapInfo[mapType]
@@ -143,6 +198,8 @@ if __name__ == "__main__":
     parser.add_argument("-f", "--features_file", dest="features_file", help="Feature file to plot", metavar="FILE", required=True)
     parser.add_argument("-o", "--features_plot", dest="features_plotname", help="Feature plot filename", metavar="FILE")
     parser.add_argument("-m", "--map_type", dest="map_type", help="The map type on which to project", metavar="FILE")
+    parser.add_argument("--max_length", dest="max_length", type=float, default=4.0,
+                        help="Maximum allowed segment length after subdivision (0.0 indicates skip subdivision)")
 
     args = parser.parse_args()
 
@@ -171,6 +228,6 @@ if __name__ == "__main__":
             plotFileName = '%s_%s.png'%(os.path.splitext(args.features_plotname)[0],mapType)
         mapInfo[mapType] = (ax, projection, plotFileName, fig)
 
-    plot_features_file(args.features_file, mapInfo)
+    plot_features_file(args.features_file, mapInfo, max_length=args.max_length)
 
 # vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/plot_features.py
+++ b/plot_features.py
@@ -15,7 +15,7 @@ all maps are used.  Possible map types are 'cyl', 'merc', 'mill', 'mill2',
 'southpole', 'atlantic', 'pacific', 'americas', 'asia'
 
 Authors: Xylar Asay-Davis, Doug Jacobsen, Phillip J. Wolfram
-Last Modified: 02/08/2016
+Last Modified: 09/14/2016
 """
 
 import os.path
@@ -24,6 +24,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 import cartopy.crs
 import cartopy.feature
+import shapely
 
 def build_projections(): #{{{
     projections = {}
@@ -50,8 +51,10 @@ def plot_base(mapType, projection): #{{{
 
     ax = plt.axes(projection=projection)
     resolution = '50m'
+
     ax.add_feature(cartopy.feature.NaturalEarthFeature('physical', 'land', resolution,
                    edgecolor='face', facecolor=cartopy.feature.COLORS['land']), zorder=1)
+
     ax.add_feature(cartopy.feature.NaturalEarthFeature('physical', 'coastline', resolution,
                    edgecolor='black', facecolor='none'), zorder=2)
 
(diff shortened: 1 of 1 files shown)


xylar:
@pwolfram, I wonder if you would consider this PR instead of #45.  As I mentioned, I found out that cartopy has an `add_geometries` function that does a lot of what we want to do with descartes but is much simpler.

Also, I have a solution for setting the bounds of the plot to be the bounds of the geometric objects being plotted: I keep a bounding box of everything as we add it to the plot and use this (transformed to each projection) to set the x and y limits.

Finally, I have separated `edgecolor` and `facecolor`, since they are different depending on the geometry type.  This means that `LineStrings` and `MutliLineStrings` don't get plotted as filled polygons. 

I was thinking of doing a PR onto #45 but I think the changes are substantial enough to be a replacement PR instead.

xylar:
# Testing

I have done the following tests so far:
```
./merge_features.py -d ocean -t Mediterranean_Basin -o Med.geojson
./plot_features.py -f Med.geojson
```
This test pulls in both transects and regions.  Some example results:

med_ortho
med_cyl

I also ran:
```
./driver_scripts/setup_MOC_basins.py --plot
./plot_features.py -f Atlantic_MOC.geojson
```
Here are some example results.  Note that the bounds are reasonable and the islands are not being filled in:
atlantic_moc
indian_moc
indopacific_moc
pacific_moc
atlantic_moc_atlantic
atlantic_moc_moll

xylar:
This PR addresses #43.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
-                plot_point(mapInfo,coords,markers[marker_num],colors[color_num])
-            else:
-                assert False, 'Geometry %s not known.'%(polytype)
-        except:
-            print 'Error plotting %s'%(feature)
+            ax.add_geometries((shape,), crs=refProjection, **props)
 
         feature_num = feature_num + 1
 
+    box = shapely.geometry.box(bounds[0],bounds[1],bounds[2],bounds[3])
```
pwolfram: I think you can more simply use the splat operator

```python
box = shapely.geometry.box(*bounds)
```

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
         (ax, projection, plotFileName, fig) = mapInfo[mapType]
         print 'saving ' + plotFileName
+
+        boxProjected = projection.project_geometry(box, src_crs=refProjection)
+        try:
+            x1, y1, x2, y2 = boxProjected.bounds
+            ax.set_xlim([x1, x2])
+            ax.set_ylim([y1, y2])
+        except ValueError:
+            # something wrong with box projection?
```
pwolfram: I would prefer that you print an error here that lets the user know something went wrong.  A better way to handle this may be to include the original error if a particular error isn't precisely found, e.g., something like https://github.com/MPAS-Dev/geometric_features/pull/45/files#diff-4069de3b3232a043d4550a22b8385ebcR175.  The reason I bring this up is because try and except blocks are really hard to debug if the python raised error isn't printed out to the user.  This can result in "silent fail" type problems.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
-                    plot_poly(mapInfo,poly,colors[color_num])
-            elif polytype == 'LineString':
-                plot_poly(mapInfo,coords,colors[color_num],filled=False)
-            elif polytype == 'Point':
-                plot_point(mapInfo,coords,markers[marker_num],colors[color_num])
-            else:
-                assert False, 'Geometry %s not known.'%(polytype)
-        except:
-            print 'Error plotting %s'%(feature)
+            ax.add_geometries((shape,), crs=refProjection, **props)
```
pwolfram: Awesome!!!

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
+            props['marker'] = marker
+
+        if bounds is None:
+            bounds = list(shape.bounds)
+        else:
+            # expand the bounding box
+            bounds[0] = min(bounds[0],shape.bounds[0])
+            bounds[1] = min(bounds[1],shape.bounds[1])
+            bounds[2] = max(bounds[2],shape.bounds[2])
+            bounds[3] = max(bounds[3],shape.bounds[3])
```
pwolfram: I would make this simpler via an element-wise minimum (http://docs.scipy.org/doc/numpy/reference/generated/numpy.minimum.html):

```python
bounds = np.minimum(bounds, shape.bounds)
```

plot_features.py
xylar: Good to know about.

plot_features.py
```diff
@@ -65,63 +68,6 @@ def plot_base(mapType, projection): #{{{
 
     return (ax,projection) #}}}
 
-def divide_poly_segments(points): #{{{
```
pwolfram: It is really nice to be able to remove functions for simplicity!

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
+        featurename = feature['properties']['name']
+        print '  feature: %s'%featurename
+
+        refProjection = cartopy.crs.PlateCarree()
+
+        color = colors[feature_num % len(colors)]
+        linestyle = linestyle[feature_num % len(linestyle)]
+        marker = markers[feature_num % len(markers)]
+
+        props = {'linewidth':2.0, 'edgecolor':color}
```
pwolfram: This is a great way to do this assignment.

plot_features.py
xylar: @pwolfram, how do I do this when 2 are mins and 2 are maxes?  I don't think an element-wise min on 2 elements and an element-wise max on 2 is really more readable.  I had it that way before and it was just bug-prone.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
+
     for mapType in mapInfo:
         (ax, projection, plotFileName, fig) = mapInfo[mapType]
         print 'saving ' + plotFileName
+
+        boxProjected = projection.project_geometry(box, src_crs=refProjection)
+        try:
+            x1, y1, x2, y2 = boxProjected.bounds
+            ax.set_xlim([x1, x2])
+            ax.set_ylim([y1, y2])
```
pwolfram: If you wanted to remove a line of code I think you could do something like

```python
ax.set_xlim(*boxProjected.bounds[0,2])
ax.set_ylim(*boxProjected.bounds[1,3])
```

However, this makes things a little less clear and you may not want to make this change.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
     # use colorbrewer qualitative 7 data class colors, "7-class Accent": http://colorbrewer2.org/
     colors = ['#7fc97f' ,'#beaed4', '#fdc086', '#ffff99','#386cb0','#f0027f','#bf5b17']
     markers = ['o', 's', 'v', '^', '>', '<', '*', 'p', 'D', 'h']
+    linestyle = ['-', '--', '-.', ':']
```
pwolfram: Do we need different linestyles?  I had originally put this in but on further reflection it may be a bad idea because it could make seeing transects harder.

plot_features.py
xylar: @pwolfram, what you say makes sense.  I'll just remove the try/except here.  I figured if the projection failed it wasn't a big deal to default back to a global projection but I agree that there should be a warning at a minimum.

I will double check but I think the reason I was getting value errors was a bug in the code.  So if I don't get the conditions where boxProjected.bounds is empty with any cases I try, I'll just assume it only happens when something is going wrong and it should raise an exception.

plot_features.py
pwolfram: You can also do the same for dictionaries, e.g, `**{'a':1,'b':2}` expands to effectively `a=1, b=2` which can be used as function arguments.

plot_features.py
pwolfram: @xylar, thanks for catching that error.  The line savings is not as great and its utility now less clear but here is the option:

```python
bounds[:2] = np.minimum(bounds[:2], shape.bounds[:2])
bounds[2:] = np.maximum(bounds[2:], shape.bounds[2:])
```

I think this is better but it is now more a matter of style so either way is ok with me.  However, if we can reduce 4 lines of code to 1 I would say that is a huge win in general.

pwolfram: I should also note that I'm not a fan of using standard python `min` and `max` and prefer the numpy versions to be used instead in general (e.g., `np.min` and `np.max`) because these are better suited to numerical computations for a variety of reasons.

xylar: Okay, I changed it as you suggested.

plot_features.py
xylar: I think that makes things less clear so I prefer to leave it as is.

plot_features.py
xylar: I don't think it's a good idea actually.  I'll remove them.

plot_features.py
xylar: I tried to remove the try/except but it causes issues when the feature just doesn't show up in the projection.  I added a warning instead.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
         print 'saving ' + plotFileName
+
+        boxProjected = projection.project_geometry(box, src_crs=refProjection)
+        try:
+            x1, y1, x2, y2 = boxProjected.bounds
+            ax.set_xlim([x1, x2])
+            ax.set_ylim([y1, y2])
+        except ValueError:
+            print "Warning: bounding box could not be projected into projection", mapType
+            print "Defaulting to global bounds."
```
pwolfram: Do you need to explicitly specify the defaulting to global bounds, e.g., `ax.set_global()`?  It may be good to do this just in case.

plot_features.py
```diff
@@ -132,40 +78,66 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
+            # expand the bounding box
+            bounds[0] = min(bounds[0],shape.bounds[0])
+            bounds[1] = min(bounds[1],shape.bounds[1])
+            bounds[2] = max(bounds[2],shape.bounds[2])
+            bounds[3] = max(bounds[3],shape.bounds[3])
 
-        color_num = feature_num % len(colors)
-        marker_num = feature_num % len(markers)
+        for mapType in mapInfo:
+            (ax, projection, plotFileName, fig) = mapInfo[mapType]
```
pwolfram: I don't think you need this line here because it will look cleaner without it, i.e., there are only two lines in the loop.

xylar: @pwolfram, ax is needed for the call below:
```
ax.add_geometries((shape,), crs=refProjection, **props)
```
I don't see a good way to eliminate the loop or the line.

plot_features.py
xylar: I will do this because I think it makes things clearer.  However, this seems to happen anyway by default.

xylar:
This PR largely addresses #46.  Other geometric types such as `GeometricCollection` could be added with relative ease with the exception that either any polygons they contain would be unfilled or any LineStrings would be filled.  It does not seem obvious how to handle more general geometric types without breaking them into constituent geometric objects.  Given that other scripts in this repository (e.g. utils) support only the geometric types currently supported by the plotting script, I think the plotting is sufficiently general.

plot_features.py
pwolfram: Sorry @xylar, I just meant the empty line wasn't needed.

pwolfram:
@xylar, can you please let me know when you are finished with changes?

xylar:
Yep, done with changes.  Feel free to merge when you think it's ready.

plot_features.py
xylar: Ah, I see.  I've removed it.

plot_features.py
```diff
@@ -135,37 +81,58 @@ def plot_features_file(featurefile, mapInfo): #{{{
...
-            print 'Error plotting %s'%(feature)
+        for mapType in mapInfo:
+            (ax, projection, plotFileName, fig) = mapInfo[mapType]
+            ax.add_geometries((shape,), crs=refProjection, **props)
 
         feature_num = feature_num + 1
 
+    box = shapely.geometry.box(*bounds)
+
+
```
pwolfram: Extra whitespace line not needed

pwolfram:
@xylar, changes look good to me except for that extra whitespace line.  I'll fully test and merge once you remove it or I can just delete that line for you-- your call.

plot_features.py
xylar: @pwolfram, this seems super picky to me but I'll change it...

xylar: Done!

pwolfram:
Thanks Xylar, feel free to push back on things like this if you want future.  Things like this aren't a big deal but from my experience this use of white space is more “standard”.  Thanks, Phillip

On Sep 15, 2016, at 8:57 AM, Xylar Asay-Davis <notifications@github.com<mailto:notifications@github.com>> wrote:


@xylar commented on this pull request.

________________________________

In plot_features.py<https://github.com/MPAS-Dev/geometric_features/pull/48>:

>
         feature_num = feature_num + 1

+    box = shapely.geometry.box(*bounds)
+
+


@pwolfram<https://github.com/pwolfram>, this seems super picky to me but I'll change it...

pwolfram:
@xylar, I'm seeing two issues with `IndoPacific_Basin.geojson`: https://gist.github.com/cf829feb9c34fd726b5e5afd9bccd3e9

First,
screenshot 2016-09-15 09 20 05

If I understand correctly, the seam should be consistent and nearly along a long of constant longitude, e.g., like

screenshot 2016-09-15 09 25 45

This seems like a bug, potentially with cartopy.

Second, `IndoPacific_Basin_robin.png`, `IndoPacific_Basin_mill2.png`, `IndoPacific_Basin_moll2.png` are just vertical lines-- do you know of a way to fix this?  Alternatively, we could just not save these plots (which may be better anyway).

I don't think either of these is a show-stopper for this PR because it is such a huge improvement but just wanted to point this out because these look like bugs.  If we merge this now we should open issues for these.

Please let me know how you would like to proceed @xylar.

xylar:
@pwolfram, the first bad seam is a known bug in cartopy (I don't know the issue hash of the top of my head) where the refinement of long line segments (in the lat/lon projection) isn't sufficiently detailed when going to curvilinear projections.  This is why we had special code in our plot script before to break segments into pieces that were no more than 4 degrees long.  You took this out in #45 because we weren't looking at the geometry at all anymore and I followed your example in this PR, and removed the unused function.  We could consider putting it back in and using it again.

xylar:
I'll take a look at the second issue.  I hadn't done those particular plotting combinations in my testing.  Sorry about that.

xylar:
@pwolfram, the second issue happens because the bounds in lon space of that particular basin are  (-180, 180), which maps to the same point in a number of projections.

I think the solution to both bugs is to put back the function that refines line segments into smaller pieces and to apply this to the bounding-box polygon as well as to all shapes.

I think this fix should be part of this PR, so I'm glad you noticed this.  I'll work on it tonight if I have time.

pwolfram:
@xylar, agreed `divide_poly_segments` needs to be included.  Will you do this prior to building the shapely instance?  I think this would be the most robust way of doing things so that plotting follows the simple cartopy api-- but this is up for debate.

xylar:
@pwolfram, yes, if I can't figure out a way to do this in shapely itself I will do it first before constructing the shapely object.

xylar:
I think it should be doable in shapely, which would be better.  Stay tuned...

xylar:
@pwolfram:
> I'm seeing two issues with IndoPacific_Basin.geojson
I think I have fixed both of these issues by subdividing the geometry before plotting it.

Here is `IndoPacific_MOC_ortho.png` with subdivision:
indopacific_moc_ortho
The seam at the anti-meridian seems to be correct now.

And here are the `mill2`, `moll2` and `robin2` plots, now that I am applying subdivision to the bounding box:
indopacific_moc_mill2
indopacific_moc_moll2
indopacific_moc_robin2

I also tested with all ocean features with the tag `Mediterranean_Basin` (which includes both transects and regions).  This also plotted correctly (as it had before), suggesting that Polygons, and LineStrings are all handled correctly.  The MOC features are MultiPolygons so those seem to work fine, too.  I combined all ocean transects into a single feature, a MultiLineString, and verified that this, too, plots correctly in all projections.

As far as I am concerned, this PR is ready to merge now.  Please perform further testing and give me a list of any stylistic changes you would like me to make before merging.

xylar:
@pwolfram, yes, I agree.  Let's do a "clean up" commit later on that applies to all the scripts.  Feel free to merge.

pwolfram:
Thanks @xylar for a truly robust PR.

PR #55, opened 2016-09-29
Remove 360-degree shift from fix_features_at_prime_meridian.py

xylar:
This merge removes an unnecessary shift by 360 degrees in the script fix_features_at_prime_meridian.py. This behavior was added for compatibility with the 0-360 degree longitude coordinate used in MPAS-O but is no longer needed because the mask creator supports features between -180 and 180 degrees longitude by default.

### DIFF
--- a/fix_features_at_antimeridian.py
+++ b/fix_features_at_antimeridian.py
@@ -4,9 +4,9 @@
 
 This script is used to split features that cross the antimeridian (+/-180 lon).
 The script reads from a single files containing the collection of features
-to be split at the antimeridian (the file can be 'features.geojson')
-and produces a file 'features.geojson'where all features are made up of
-polygons that do not cross the antimeridian.
+to be split at the antimeridian (the file can be the same as the output file)
+and produces a file indicated by the -o flag ('features.geojson' by default
+where all features are made up of polygons that do not cross the antimeridian.
 
 The code maps features into polar coordinates centered at the closest pole,
 which implicitly assumes that no segment of a feature will cover more than
--- a/fix_features_at_prime_meridian.py
+++ b/fix_features_at_prime_meridian.py
@@ -2,13 +2,14 @@
 
 """
 
-This script is used to split features across the prime meridian.  The 
+This script is used to split features across the prime meridian.  The
 features should not cross the antimeridian (+/-180 lon), so the user should
 call fix_features_at_antimeridian.py first if necessary before using this
-tool.  The script reads from a single files containing the collection of 
-features to be split at the prime meridian (the file can be 'features.geojson')
-and produces a file 'features.geojson' where all features are made up of
-polygons that do not cross the prime meridian.
+tool.  The script reads from a single files containing the collection of
+features to be split at the prime meridian (the file can be the same as the
+output file) and produces a file indicated by the -o flag ('features.geojson'
+by default) where all features are made up of polygons that do not cross the 
+prime meridian.
 
 The script makes use of the shapely library.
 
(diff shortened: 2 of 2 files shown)


xylar:
@pwolfram, this should be merged after #52.  It is a relatively simple bug fix.

xylar:
I have tested both `fix_features*.py` scripts on the following features file to make sure it still works as expected.

```
{"type": "FeatureCollection",
 "features":
	[
		{"type": "Feature",
		 "properties": {
			"author": "Xylar Asay-Davis",
			"component": "ocean",
			"name": "temp",
			"object": "region",
			"tags": ""
		 },
		 "geometry":
			{"type": "Polygon",
			 "coordinates":
				[
					[
						[ 180.000000, -85.000000],
						[ 90.000000, -85.000000],
						[ 0.000000, -85.000000],
						[ -90.000000, -85.000000],
						[ -180.000000, -85.000000],
						[ 180.000000, -85.000000]
					]
				]
			}
		}
	]
}
```
The result of the following commands:
```
./fix_features_at_antimeridian.py -f 85S.geojson -o valid_85S.geojson
./fix_features_at_prime_meridian.py -f valid_85S.geojson -o valid_85S_split_at_prime.geojson 
```
is a shape split at both meridians, as expected:
```
{
    "type": "FeatureCollection", 
    "features": [
        {
            "type": "Feature", 
            "properties": {
                "author": "Xylar Asay-Davis", 
                "object": "region", 
                "component": "ocean", 
                "name": "temp", 
                "tags": ""
            }, 
            "geometry": {
                "type": "MultiPolygon", 
                "coordinates": [
                    [
                        [
                            [
                                -180.000000, 
                                -85.000000
                            ], 
                            [
                                -90.000000, 
                                -85.000000
                            ], 
                            [
                                0.000000, 
                                -85.000000
                            ], 
                            [
                                0.000000, 
                                -90.000000
                            ], 
                            [
                                -180.000000, 
                                -90.000000
                            ], 
                            [
                                -180.000000, 
                                -85.000000
                            ]
                        ]
                    ], 
                    [
                        [
                            [
                                0.000000, 
                                -85.000000
                            ], 
                            [
                                90.000000, 
                                -85.000000
                            ], 
                            [
                                180.000000, 
                                -85.000000
                            ], 
                            [
                                180.000000, 
                                -90.000000
                            ], 
                            [
                                0.000000, 
                                -90.000000
                            ], 
                            [
                                0.000000, 
                                -85.000000
                            ]
                        ]
                    ]
                ]
            }
        }
    ]
}
```

pwolfram:
@xylar, can you please rebase this branch and then I'll review it?

xylar:
@pwolfram, in the end, I don't think the rebase was necessary because I think a merge would have detected the redundant commits.  But I just did the rebase in any case.  Removing the don't merge flag.

pwolfram:
@xylar, this looks good.  Some thoughts:

1. The functionality between both of these scripts is very similar.  Do we need two scripts?
2. The script `fix_features_at_antimeridian.py` never appears to be used in the workflow:
```bash
┌─[pwolfram][shapiro][~/Documents/MPAS_pull_requests/fix_feature_prime_meridian_shift][11:26][±][fix_fix_prime_meridian_scrip ✗]
└─▪ git grep fix_features
driver_scripts/setup_antarctic_basins.py:41:args = ['./fix_features_at_prime_meridian.py', '-f', outName]
driver_scripts/setup_ocean_land_coverage.py:46:#args = ['./fix_features_at_prime_meridian.py', '-f', outName]
fix_features_at_prime_meridian.py:7:call fix_features_at_antimeridian.py first if necessary before using this
```
I'm happy to merge this but it may be useful to consider these potential cleanup steps, which are obviously not high-priority.

xylar:
@pwolfram, I think there is a logical difference between the two scripts and we had rather different uses for them.  In fact, we may not need `fix_features_at_prime_meridian.py` anymore even though it's currently used on the Antarctic basins.

`fix_features_at_antimeridian.py` is used to fix features that are not valid because they contain the North and/or South poles or because they cross the antimeridian and thus have segments that erroneously connect points near -180 with those near 180.  Features like these arose when making Antarctic basins on polar stereographic grids and then remapping the points onto lon/lat coordinates.  One basin included the South pole and several crossed the -180/180 divide.  This script was applied manually to all Antarctic features before committing them to the repo.  If we had provenance for those features, this script would appear there.  Unfortunately, we don't and we'll need to recreate it.  I'll add that to my to-do list.

`fix_features_at_prime_meridian.py` was used to make features that the MPAS mask creator could read with longitude between 0 and 360 degrees.  This is no longer needed because the mask creator assumes it's getting input between -180 and 180 by default now.

pwolfram:
@xylar, should we remove `fix_features_at_prime_meridian.py` then?  I'm ok leaving it for now but want to keep the repository as simple / clean as possible (hence the suggestion to potentially remove it).

xylar:
@pwolfram, you could make an issue for this.  For now, I'm not entirely sure it's not needed so it should stay.

Issue #62, opened 2016-09-29
Provenance for ocean, ice shelf and land ice regions


xylar:
Scripts should be added to `feature_creation_scripts` that can recreate all the features in the repository that weren't created by hand.

xylar:
@stephenprice, if you can give me whatever scripts you used to create the Antarctic features at some point, I can see that they make their way to the repo so we can thereby document how we created the features we're using.  I'll do the same for ocean and ice shelf features as I find time.

pwolfram:
@xylar, should we additionally be somehow documenting the workflow used to produce each derivative feature via a "history" property that records the commands and git hash used to produce the feature?

xylar:
@pwolfram, I guess that's probably right.  I don't know how to do that effectively so maybe you should make it a separate issue or design document or whatever you think is appropriate.

xylar:
Addressed by #64

PR #64, opened 2016-09-29
Provenances features via new history property

pwolfram:
This merge provenances features files, e.g.,

``` bash
$ ./tag_features.py -f ocean/region/North_Atlantic_Ocean/region.geojson -t test -o tester.geojson
$ head -n 12 tester.geojson
{
    "type": "FeatureCollection", 
    "features": [
        {
            "type": "Feature", 
            "properties": {
                "name": "North Atlantic Ocean", 
                "tags": "North_Atlantic_Ocean;Atlantic_Basin;test", 
                "object": "region", 
                "component": "ocean", 
                "author": "http://www.marineregions.org/downloads.php#iho", 
                "history": "09/29/16 15:47 : hostname : pwolfram : /Users/pwolfram/Documents/MPAS_pull_requests/provenance_features : 2bf4bf8 : ./tag_features.py -f ocean/region/North_Atlantic_Ocean/region.geojson -t test -o tester.geojson;"
```

This will better keep track of operations performed on features files.

cc @xylar and #62

### DIFF
--- a/utils/feature_write_utils.py
+++ b/utils/feature_write_utils.py
@@ -3,18 +3,20 @@
 """
 Utility funcitons for writing geojson files from a dictionary of features.
 
-Authors: Douglas Jacobsen, Xylar Asay-Davis
-Last Modified: 10/22/2016
+Authors: Douglas Jacobsen, Xylar Asay-Davis, Phillip J. Wolfram
+Last Modified: 03/16/2017
 """
 
 import json
 import sys
+from ioprovenance import provenance_command
 
 from collections import OrderedDict
 
 
 def write_all_features(features, out_file_name, indent=4,
-                       defaultGroupName='enterGroupName'):  # {{{
+                       defaultGroupName='enterGroupName',
+                       strip_history=False):  # {{{
     json.encoder.FLOAT_REPR = lambda o: format(o, 'f')
 
     for index in range(len(features['features'])):
@@ -38,6 +40,21 @@ def write_all_features(features, out_file_name, indent=4,
 
     # features go last for readability
     outFeatures['features'] = features['features']
+
+    if strip_history:
+        # remove provenance from the output
+        for feature in outFeatures['features']:
+            # pop (with default so no exception is raised if no history)
+            feature['properties'].pop('history', None)
+    else:
+        # provenance the output
+        command = provenance_command()
+        for feature in outFeatures['features']:
+            if 'history' in feature['properties']:
+                appendcmd = feature['properties']['history'] + ' ' + command
+                feature['properties']['history'] = appendcmd
+            else:
+                feature['properties']['history'] = command
 
     out_file = open(out_file_name, 'w')
 
--- a/utils/ioprovenance.py
+++ b/utils/ioprovenance.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+"""
+Utility funcitons for provenance of feature files.
+
+Authors: Phillip J. Wolfram
+Last Modified: 09/29/2016
+"""
+
+import os
+import sys
+import socket
+import subprocess
+import numpy as np
+import datetime
+
+def provenance_command(): #{{{
+    cwd = os.getcwd()
+    user = os.getenv('USER')
+    curtime = datetime.datetime.now().strftime('%m/%d/%y %H:%M')
+    call = ' '.join(sys.argv)
+    host = socket.gethostname()
+    try:
+        githash = subprocess.check_output(['git', 'describe', '--always','--dirty']).strip('\n')
+    except:
+        githash = 'None'
+    sep = ' : '
+    provstr = sep.join([curtime,  host, user, cwd, githash, call]) + ';'
+    return provstr #}}}
+
+if __name__ == "__main__":
+    print provenance_command()
+
+# vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python
--- a/split_features.py
+++ b/split_features.py
@@ -58,6 +58,6 @@ for feature in features_file['features']:
                                                  dir_name, object_type)
 
     write_all_features({'features': [feature]}, out_file_name, indent=4,
-                       defaultGroupName=None)
+                       defaultGroupName=None, strip_history=True)
 
 # vim: foldmethod=marker ai ts=4 sts=4 et sw=4 ft=python


pwolfram:
@xylar, you can easily see the form history property that is added via 
```bash
└─▪ python utils/ioprovenance.py
09/29/16 15:51 : hostname : pwolfram : /Users/pwolfram/Documents/MPAS_pull_requests/provenance_features : c4a7a57 : utils/ioprovenance.py;
```

xylar:
@pwolfram, this looks really good but I likely won't have a chance to review it until mid October.

pwolfram:
No problem @xylar, I hope you have a nice trip.

pwolfram:
@milenaveneziani, could you please take a look at this when you get the chance?  The goal here is to make sure each region file is at least partially self-documenting.  Thanks!

pwolfram:
@milenaveneziani, this would be good to help with #70 to make sure that we know precisely how each file is produced.

milenaveneziani:
Looks good to me, @pwolfram.

utils/feature_write_utils.py
```diff
@@ -33,6 +34,15 @@ def write_all_features(features, out_file_name, indent=4):  # {{{
         if key not in outFeatures.keys():
             outFeatures[key] = features[key]
 
+    # provenance the output
+    command = provenance_command()
+    for afeature in features['features']:
```
xylar: @pwolfram, out of curiosity, why `afeature` and not just `feature`?

utils/feature_write_utils.py
```diff
@@ -33,6 +34,15 @@ def write_all_features(features, out_file_name, indent=4):  # {{{
...
+        if afeature['properties'].get('history', False):
```
xylar: My understanding is that the preferable way to write this would be:
```python
if 'history' in afeature['properties']:
```

utils/ioprovenance.py
```diff
@@ -0,0 +1,33 @@
...
+
+import os
+import sys
+import socket
+import subprocess
+import numpy as np
+import datetime
+
+def provenance_command(): #{{{
+    cwd = os.getcwd()
```
xylar: I have minor privacy concerns about `cwd` being included in the history.  Do you feel that the usefulness of this info (which I perceive as being rather short lived) outweighs the fact that my directory structure becomes publicly available when I create geometric features?

xylar:
@pwolfram, I wonder if we want history to be stripped out by or perhaps made less verbose in the `split_features.py` script.  The concern I have is that various pieces of information that are useful to include in features that we control locally are not necessarily appropriate for sharing publicly via GitHub.  I don't necessarily want people to know which machine I used to create a geometric feature in the repo, what the directory structure looks like and what my username is.  Maybe I'm just paranoid that way...

xylar:
@pwolfram, another issue with including history for files written out with `split_features.py` is that the current behavior, which is very much desired, is that `merge_features.py` followed immediately by `split_features.py` should leave the files in the repo completely unchanged.  This behavior is very useful for determining which features have been modified by a given process (e.g. masking features) and for preventing unnecessary updates to the repository when nothing useful has changed in a feature.

On the other hand, I can see value in having provenance for all the features in the repo that is more focused on which commands were used to create them and less on which machine, user, etc. were involved.

xylar:
@pwolfram, I wonder if a solution to my concern might be to have a shorter history attribute for each feature and a longer history attribute for the feature collection.  Here are the advantages I see:
  1. The history of individual features could be restricted to the command and the `geometric_features` git hash, and perhaps other relevant information for their reconstruction.  This would leave out machine, username, path, etc. that raises security/privacy concerns for me.
  2. The history attribute for the feature collection could have machine, username, path, etc. so that information could be used to track the more complete history of a given file, e.g. within the MPAS and ACME teams.
  3. The history from feature collections would be stripped out by `split_features.py`, since this process conceptually breaks up the feature collection into its constituents.  We could still discuss what happens to the history attribute for individual features when `split_features.py` is invoked.  On the one hand, it is useful if their history is preserved so they could be reconstructed if need be.  On the other hand, it would not be particularly useful to preserve their history if the history itself is the only thing that has changed (e.g. if one calls `merge_features.py` followed by `split_features.py`).
  4. The history of the feature collection would be preserved even when invoking operations like `combine_features.py` that do not preserve the indivudual features (and thus their histories).

pwolfram:
@xylar, here is an interesting idea: we could have an option that stores a non-globally accessible hash name, e.g., you'd store a name you could search in your private gists that has the key private information needed to fully reconstruct the workflow.  This obviously wouldn't work everywhere but could be a solution if needed.  However, this is probably overkill.

Regarding 1 and 2 I'm a little unclear on what information should be blacklisted, e.g., is it only username and path?  Having a blacklist for privacy concerns would be helpful to make sure this information is not included in the provenance.

xylar:
@pwolfram wrote:
> we could have an option that stores a non-globally accessible hash name, e.g., you'd store a name you could search in your private gists that has the key private information needed to fully reconstruct the workflow. 

Yep, I think that's overkill.

> Regarding 1 and 2 I'm a little unclear on what information should be blacklisted, e.g., is it only username and path? Having a blacklist for privacy concerns would be helpful to make sure this information is not included in the provenance.

I'd prefer that individual features not have username and path, and I don't think machine name is useful without those two, so I would leave that out, too.

xylar:
@pwolfram, let me know if this is something you're interested in revisiting.  I think we had reached a consensus that what was needed was a way to remove the `history` property from each feature as part of `split` so that 1) combining features and then splitting them again didn't change the original features and 2) there wasn't potentially sensitive information stored in the public features (user names, machine names, directories, etc.)

I think that would be a relatively simple addition to this PR and then we could probably merge it.  What do you think, should I look into that or would you like to?

xylar:
@pwolfram, this is the only outstanding PR on my plate and I'd like to address it if we can.  I realize you're busy so I've made a PR to your branch with the changes I'd like to see.  If you're okay with those changes and merge that PR, I can take it from there.  If you want to (and have time to) resolve the conflicts with `master`, that's fine.  If not, I'll take care of it.

pwolfram:
Note, CI fails because we haven't merged #88 yet...

utils/ioprovenance.py
pwolfram: Did you think we should remove this provenance?  Also, do we need a history scrubber script too in this PR to help alleviate privacy concerns?

xylar: Now that it gets stripped out during `split_features.py`, I have no concerns.

xylar: I don't think we need a scrubber.  People can use their best judgment about what to make public, but I now see how much of this stuff is in all the NetCDF files we have around so I guess that ship has definitely sailed.

pwolfram:
Thanks!

PR #67, opened 2016-10-05
Add groupName as an attribute to merged feature files

### LINKED ISSUE #65: groupName removed from geojson files when running setup_MOC_basins and setup_ocean_basins

douglasjacobsen:
This merge adds groupName to the merged feature files, to allow feature
groups to be created by subsequent tools.

Fixes #65

### DIFF
--- a/combine_features.py
+++ b/combine_features.py
@@ -32,6 +32,9 @@ parser.add_argument("-f", "--feature_file", dest="feature_file",
 parser.add_argument("-n", "--new_feature_name", dest="new_feature_name",
                     help="The new name of the combined feature",
                     metavar="NAME", required=True)
+parser.add_argument("-g", "--groupName", dest="groupName",
+                    help="Feature group name.",
+                    metavar="GROUPNAME", default="unspecifiedGroupName")
 parser.add_argument("-o", "--output", dest="output_file_name",
                     help="Output file, e.g., features.geojson.",
                     metavar="PATH", default="features.geojson")
@@ -97,6 +100,7 @@ feature['properties']['author'] = '; '.join(list(set(authors)))
 feature['properties']['constituents'] = '; '.join(list(set(featureNames)))
 feature['geometry'] = shapely.geometry.mapping(combinedShape)
 features['features'].append(feature)
+features['groupName'] = args.groupName
 
 if feature['geometry']['type'] == 'GeometryCollection':
     print "Error: combined geometry from %s is of type GeometryCollection."%(args.feature_file)
--- a/difference_features.py
+++ b/difference_features.py
@@ -31,6 +31,9 @@ parser.add_argument("-m", "--mask_file", dest="mask_file",
                     help="Feature file with one or more features whose overlap "
                          "with features in feature_file should be removed",
                     metavar="FILE2", required=True)
+parser.add_argument("-g", "--groupName", dest="groupName",
+                    help="Feature group name.",
+                    metavar="GROUPNAME", default="unspecifiedGroupName")
 parser.add_argument("-o", "--output", dest="output_file_name",
                     help="Output file, e.g., features.geojson.",
                     metavar="PATH", default="features.geojson")
@@ -107,6 +110,8 @@ for feature in featuresToMask['features']:
         features['features'].append(feature)
     else:
         print "%s has been removed."%name
+
+features['groupName'] = args.groupName
 
 write_all_features(features, out_file_name, indent=4)
 
--- a/driver_scripts/setup_MOC_basins.py
+++ b/driver_scripts/setup_MOC_basins.py
@@ -19,6 +19,8 @@ parser.add_option("--plot", action="store_true", dest="plot")
 
 options, args = parser.parse_args()
 
+groupNameMOC = 'MOCBasinRegionsGroup'
+groupNameBasins = 'OceanBasinRegionsMOCGroup'
 
 subBasins = {'Atlantic': ['Atlantic', 'Mediterranean'],
              'IndoPacific': ['Pacific', 'Indian'],
@@ -52,11 +54,11 @@ for basinName in subBasins:
     #merge the the features into a single file
     print " * combining features into single feature named %s_MOC"%basinName
     spcall(['./combine_features.py', '-f', basinFileName, '-n', '%s_MOC'%basinName,
-            '-o', basinCombinedFileName])
+            '-g', groupNameBasins, '-o', basinCombinedFileName])
 
     print " * masking out features south of MOC region"
     spcall(['./difference_features.py', '-f', basinCombinedFileName,
-            '-m', MOCMaskFileNames[basinName], '-o', MOCName])
+            '-m', MOCMaskFileNames[basinName], '-g', groupNameMOC, '-o', MOCName])
 
     if options.plot:
         spcall(['./plot_features.py', '-f', MOCName, '-o', imageName, '-m', 'cyl'])
--- a/driver_scripts/setup_ocean_basins.py
+++ b/driver_scripts/setup_ocean_basins.py
@@ -18,6 +18,8 @@ parser.add_option("--plot", action="store_true", dest="plot")
 
 options, args = parser.parse_args()
 
+groupName = 'OceanBasinRegionsGroup'
+
 for oceanName in 'Atlantic', 'Pacific', 'Indian', 'Arctic', 'Southern_Ocean', 'Mediterranean':
 
     tag = '%s_Basin'%oceanName
@@ -37,7 +39,7 @@ for oceanName in 'Atlantic', 'Pacific', 'Indian', 'Arctic', 'Southern_Ocean', 'M
     #merge the the features into a single file
     print " * combining features into single feature named %s_Basin"%oceanName
     spcall(['./combine_features.py', '-f', basinFileNameName, '-n', '%s_Basin'%oceanName,
-            '-o', basinCombinedFileName])
+            '-g', groupName, '-o', basinCombinedFileName])
     
     if(options.plot):
         args = ['./plot_features.py', '-f', basinCombinedFileName, '-o', imageName, '-m', 'cyl']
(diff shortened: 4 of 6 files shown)


douglasjacobsen:
See https://github.com/MPAS-Dev/geometric_features/pull/66#issuecomment-251691765 about a conversation related to the motivation / use of groupName.

Note, this doesn't update the driver scripts to automatically set unique / useful groupNames for the resulting feature group files.

pwolfram:
Note, following the group meeting today, @douglasjacobsen mentioned that this should not be as general as in this PR but instead should be included when using the merge script.

douglasjacobsen:
I just pushed an updated version of this branch that does what I think it should in terms of adding groupName to features files.

Note, with the change to json.dump, you'll have to re-merge / dump all of the existing features to prevent people from thinking the scripts have changed them (for example, if you merge any ocean region, and then split it, the entire file is different).

milenaveneziani:
Thanks @douglasjacobsen for doing this.
I think I can test this by running some driver_scripts, right?
Also, I don't think I understood the last 3 lines of Doug's comment: is it something we need to do before merging this PR?

douglasjacobsen:
@milenaveneziani You can test this with two things. 1) run a driver script, and make sure groupName is an attribute. 2) Split some features to where they are supposed to go, and make sure groupName is *not* an attribute.

You don't need to do anything else for this specific PR, I was just saying that when you do (2) above, the entire file will look like it's changed (because the formatting has changed) so it might be confusing. At some point you guys should re-merge / split all of the features within this repo, to make sure their formatting is consistent with the output format. This is one of the negative side effects of changing the output format.

milenaveneziani:
I see now. I am hoping @xylar will have a better idea of what features will be impacted by this PR.

I ran the setup_MOC_basins driver and the groupName is included in the *_Basin geojson files (which are created with merge_features), but not in the *_MOC geojson's, because those are created with combine_features.py

milenaveneziani:
Ok, so I added the groupName to the other 2 scripts that are used to create the MOC basins: combine_features and difference_features. Now, if I run setup_MOC_basins, all resulting features have the groupName in them. Same goes for setup_ocean_basins, as it should be.

I will let @xylar take a look at this though before merging.

pwolfram:
@milenaveneziani, I think we will ultimately want `"enterGroupName"` to be specified via a command line argument, e.g., `-g groupName` in each of the scripts.  I'm happy to make this change or you can, please let me know either way.

milenaveneziani:
@pwolfram: I thought of adding the -g option too. Here is my attempt. Let me know if you have comments.
I also added a specific groupName for MOC and Basins regions in setup_MOC_basins and setup_ocean_basins. Not sure if this should be made an option. I thought against it being an option to reduce degrees of freedom when writing the plotting/analysis script.

milenaveneziani:
Also, I ran split_feature on all *_Basin_separate feature files I had created with setup_ocean_basins and setup_MOC_basins, so that 1) I made sure that groupName was *not* an attribute in the resulting split features, and 2) I changed the format to most of the features in ocean/region. I ran split_features also on the remaining (11) ocean/region/geojson files that were left with the old format, so now all features on my local branch have the new format.
I am not committing these new features quite yet, though, to double check with you if this is OK.

merge_features.py
```diff
@@ -101,6 +102,10 @@
             print "Error parsing geojson file: %s"%(path)
     del paths
 
+#if ( 'groupName' not in all_features.keys() ):
+#    all_features['groupName'] = "enterGroupName"
```
pwolfram: Please remove commented text because it is not needed.

driver_scripts/setup_MOC_basins.py
```diff
@@ -19,6 +19,8 @@ def spcall(args): #{{{
 
 options, args = parser.parse_args()
 
+groupNameMOC = 'MOCRegionsGroup'
+groupNameBasins = 'BasinRegionsGroup'
```
pwolfram: I'm wondering if there isn't a choice of better name for `groupNameBasins`.  We probably want to make sure the group names are unique throughout the repo too.

combine_features.py
```diff
@@ -32,6 +32,9 @@
 parser.add_argument("-n", "--new_feature_name", dest="new_feature_name",
                     help="The new name of the combined feature",
                     metavar="NAME", required=True)
+parser.add_argument("-g", "--groupName", dest="groupName",
+                    help="Feature group name.",
+                    metavar="GROUPNAME", default="enterGroupName")
```
pwolfram: Instead of `"enterGroupName"` I think we want something like `"unspecifiedGroupName"` and this should be a default throughout the changes to this repo.

difference_features.py
```diff
@@ -31,6 +31,9 @@
                     help="Feature file with one or more features whose overlap "
                          "with features in feature_file should be removed",
                     metavar="FILE2", required=True)
+parser.add_argument("-g", "--groupName", dest="groupName",
+                    help="Feature group name.",
+                    metavar="GROUPNAME", default="enterGroupName")
```
pwolfram: Please see previous comment about changing default.

driver_scripts/setup_ocean_basins.py
```diff
@@ -18,6 +18,8 @@ def spcall(args): #{{{
 
 options, args = parser.parse_args()
 
+groupName = 'BasinRegionsGroup'
```
pwolfram: This groupName is not unique with respect to `setup_MOC_basins.py`.

merge_features.py
```diff
@@ -30,6 +30,7 @@
 parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
 parser.add_argument("-f", "--feature_file", dest="feature_file", help="Single feature file to append to features.geojson", metavar="FILE")
 parser.add_argument("-d", "--features_directory", dest="features_dir", help="Directory containing multiple feature files, each will be appended to features.geojson", metavar="PATH")
+parser.add_argument("-g", "--group", dest="groupName", help="Feature group name.", metavar="GROUPNAME", default="enterGroupName")
 parser.add_argument("-t", "--tags", dest="tags", help="Semicolon separated list of tags to match features against.", metavar='"TAG1;TAG2;TAG3"')
```
pwolfram: Please see previous comment about changing default.

pwolfram:
@milenaveneziani, I think this makes sense and if something seems like it could be improved I commented on it in the review.  Note, we need to do the 

> Note, with the change to json.dump, you'll have to re-merge / dump all of the existing features to prevent people from thinking the scripts have changed them (for example, if you merge any ocean region, and then split it, the entire file is different).

step @douglasjacobsen recommended (reprinted above for clarity).  Note this is up for debate as to whether this should be done here or in a later PR, but I think we want to make this a giant merge for clarity.  @xylar, please feel free to comment on this if you disagree.

xylar:
> What was the purpose of the change to ocean/region/Southern_Ocean/region.geojson? 

I am wondering the same.  @douglasjacobsen included this in his commit, but I'm wondering if that was by mistake and should be undone.

We have also undone @douglasjacobsen's only other change, so maybe the first commit on this branch should be rebased out.


> We eventually need to update all the files to correspond to the changes with this script but need to discuss whether this should be now or in a later PR. I'm leaning toward now as opposed to later because then all the changes in this PR are consistent and not inter-dependent on another PR but my initial reaction was the opposite. What do you think about this @milenaveneziani?

> If you agree, you should push the changes to this repo for review. However, we have already changed the output format in #52 so it wouldn't be a big deal to submit all the changes in a separate PR. In hindsight, I think @xylar and I should have updated all the geojson files when we completed #52.

I strongly feel like this needs to be a separate PR.  I really dislike when PRs creep way beyond their initial intended purpose just because there is a need for other clean up.  While I suppose this cleanup should have happened as part of #52, I don't think we split features very often at this point because most of the features we want to define are now in the repo, so I didn't see much urgency in performing this update when I posted #52.

pwolfram:
Following @xylar's comments I propose we don't change any geojson files in this repo and move that we save geojson changes to a fresh PR.

milenaveneziani:
@pwolfram and @xylar: about the Souther_Ocean geojson file change, I was also puzzled by why that happened. But I did mention in one of my comments that I have run split_features (as Doug suggested) on the *_Basin_separated files and all the remaining features, so I now have the newer format on my local branch.
If you prefer these changes to go in a different PR, could you please suggest how to undo the current change in the SO geojson file?

driver_scripts/setup_MOC_basins.py
milenaveneziani: What about OceanBasinsGroup?
I wanted to leave Regions ingroupNameMOC because at some point we will also have MOCTransectsGroup.
Also, this is the only place where we define specific groups for now.
Another 2 regions that should fall in the OceanBasinsGroup are the GlobalOcean_* features: any suggestion in whether this should be done by hand or within the setup_ocean_basins driver?

driver_scripts/setup_ocean_basins.py
pwolfram: The name sounded too generic and not specific enough.  Maybe something like `GlobalOceanBasinRegionsGroup` would be better?

driver_scripts/setup_MOC_basins.py
pwolfram: To summarize:

 1. `groupNames` should be unique from each script
 2. I would include elements like `MOC`, `Global`, etc. as well as `Regions` and `Transects` for greater specificity in `groupNames`

pwolfram:
@milenaveneziani, I'll take care of removing the SO geojson file as we discussed.

pwolfram:
@milenaveneziani, we observe the problem with `Atlantic_Basin.geojson` produced by `setup_ocean_basins.py` and `setup_MOC_basins.py` yielding different output but having the same `groupName`.  
screenshot 2016-10-06 11 45 01

The coordinates are also very different.  This is something to keep in mind when using these results in the model because I would say that having the same `groupName` point to different features is dangerous long term.  

The more I think about it we need to have a different `groupName` for each script and I would argue that this has to happen in this PR, not another one, because we are grappling with the `groupName` issue here.

pwolfram:
In summary, one of `OceanBasinRegionsGroup` needs renamed.

milenaveneziani:
@pwolfram: now this is confusing to me too.. I really thought the two Atlantic_Basin features generated by the 2 drivers were the same. I would say let us have a unique groupName for this PR, but we need to address this asap. I was just about to open an issue about this.

milenaveneziani:
@pwolfram: would you mind making the change and pushing, since you are already doing some clean up on your side? thanks.

Issue #69, opened 2016-10-05
Adding definition of MOC regions southern transects to setup_MOC_basins


milenaveneziani:
The MOC and MHT AM's need both the MOC regions and their southern border for doing their computation. Since these two features need to be consistent with each other, I thought that the driver setup_MOC_basins could do both things:
1) define the MOC ocean basins;
2) pick the southern border of each so defined basin and make a new transect feature.
Does this sound like a reasonable thing to do?

pwolfram:
@milenaveneziani, how were you planning on defining the southern border?  Were you going to specify at latitude such that the part of the polygon south of the latitude is the transect?

milenaveneziani:
@pwolfram: each MOC region is defined by a specific southern latitude. I am thinking this is all that is needed as far as geojson files go. Then the mask creator will do its job, and I don't know whether it takes the southern/northern/or closest part of the polygon wrt that latitude.

xylar:
@milenaveneziani and @pwolfram, I think 2) above would require a new tool and some new features in the repo.  The new tool would be `./intersect_features.py`, which would be the conceptual opposite of the `./difference_features.py` script: The source features would be masked to their intersection with the masking features(s) provided.  The new features would be transects running from -180 to 180 degrees at the southern boundary of each MOC region (similar to the MOC mask features but transects instead of regions).  Then, the desired southern boundary could be created as the intersections of these regions with the appropriate basin.

I will be happy to write this tool when I get back from France on the 17th.

xylar:
@milenaveneziani wrote:
> Then the mask creator will do its job, and I don't know whether it takes the southern/northern/or closest part of the polygon wrt that latitude.

For each segment of each LineString in a transect feature, the mask creator will find a set of cell centers (and I believe also the edges and vertices, but I've not used that functionality) that are closest to that segment.  I assume you are after the transport across the southern boundary of each basin?  @mark-petersen has used transects in MPAS-O to compute transport so he may have some pointers.

Have you thought about the northern boundaries at all?  They are not as clean but could theoretically also be defined by transects.  Are they safe to ignore for some reason, whereas the southern boundaries definitely matter?

mark-petersen:
@xylar, the MOC is computed as a stream function Psi(y,z) and can be computed from:
(1) dPsi/dz = V   (V is zonally averaged meridional velocity)
(2) dPsi/dy = W  (W is zonally averaged vertical velocity)
[...]

PR #75, opened 2016-10-16
Create southern boundary features for each MOC basin

xylar:
This merge adds a script than can be use to intersect features.  The features
in one file are intersected with the intersection of features in another file (which
will typically contain a single feature in any case).

This new script is used as part of the driver script to create MOC basins to
intersect two new transects, `MOC_34S` and `MOC_6S`, with the MOC basin
features to create transects defining the southern boundary of each basin.


xylar:
@milenaveneziani, this is intended to address #69.  It depends on #74, so should not be merged until that PR has been fulfilled.  But it could be helpful to know if this is working the way you need it to before #74 gets merged (assuming that might take a little time).

xylar:
# Testing

I viewed `MOCSounthernBoundaries.geojson`, which is now one of the results of `setup_MOC_basins.py`, and verified that it contains 4 features defining the southern boundaries of each of the 4 MOC basins.

I have not preformed any other testing of `intersect_features.py` so far.

milenaveneziani:
@xylar: I created MOCSounthernBoundaries.geojson using this code, and then ran it through the mask creator tool.
I received this error message at the end:

Built 714237 edges.
 Building closest edge location for line segments
 Reading edge graph
 Building edge transects
 Building edge transect path signs
 ERROR: Edges 580335 and 298234 don't share any vertices. Exiting...

xylar:
@milenaveneziani, thanks for trying this out.  I'll have to investigate further.  I don't know much about how the mask creator works so it's hard for me to guess what the problem might be.  Feel free to dig into the code yourself if you have the time and/or c++ expertise to do so.

xylar:
@milenaveneziani, I don't think the southern boundary as defined in this PR is what is needed for the MOC computation.  Instead, I think what is actually needed is the following:
  1. Small polygons should be removed from each MOC basin.  These arise because a bit of the amazon is still in the Atlantic basin, a bit of the Tasman Sea is still in the Indopacific basin, etc.  These small polygons have no connection to the southern boundary of the main basin.
  2. Once small polygons are removed, the southern boundary should consist of all points on the perimeter of the polygon(s) defining the MOC basin that lie between the west-most point at the southern-most latitude in the basin and the eastern-most point at the southern-most latitude.  For example, for the Atlantic MOC basin, the southern-most latitude is -34 degrees so we find the western-most and eastern-most points at this latitude, and define the southern boundary as the part of the perimeter of the the Atlantic MOC basin that connects these two.

Both of these steps will require tools that are more sophisticated than the basic manipulation tools currently available in this repo.  It is not yet clear to me if the tool should be included directly in `setup_ocean_basins.py` or if it should be a standalone module.

xylar:
#78 replaces this PR.

PR #71, opened 2016-10-06
Changed feature's format for all features by running split_features

milenaveneziani:
The ocean/region/*geojson features now have the newer format. They were crated by running them through split_features.py.

PR #77, opened 2016-10-21
Sort properties before writing to geojson

xylar:
feature properties are placed in an ordered dictionary before
writing.  Keys other than type and groupName in feature collections
are also sorted, as are keys other than type and properties in
individual features.

### DIFF
--- a/utils/feature_write_utils.py
+++ b/utils/feature_write_utils.py
@@ -4,7 +4,7 @@
 Utility funcitons for writing geojson files from a dictionary of features.
 
 Authors: Douglas Jacobsen, Xylar Asay-Davis
-Last Modified: 09/29/2016
+Last Modified: 10/22/2016
 """
 
 import json
@@ -19,7 +19,7 @@ def write_all_features(features, out_file_name, indent=4,
 
     for index in range(len(features['features'])):
         features['features'][index] = \
-            check_feature(features['features'][index])
+            _check_feature(features['features'][index])
 
     features['type'] = 'FeatureCollection'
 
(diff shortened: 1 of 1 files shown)


xylar:
This merge is intended to make sure geojson files remain bit-for-bit (at least to the extent possible).  By sorting properties, we are able to ensure that they will be written in the same order, previously the ordering of properties was arbitrary.

xylar:
@pwolfram, I think this PR should take care of the non-bit-for-bit issue you were seeing.  Let me know if not.

xylar:
@pwolfram, we could also make the geometry dictionary be an `OrderedDict` for completeness if you like.  It wouldn't be hard.  So far I haven't seen cases where it's an issue but there's no guarantee of bit-for-bit if we don't.  I guess I'll just go ahead and change it.

pwolfram:
@xylar, could we please change `check_feature` to `_check_feature` to indicate it is a private function for this module?  As far as I can tell there is no planned usage for this function outside this module and it would make the code a little clearer.

pwolfram:
Otherwise, this PR is great and I'm happy to merge.

xylar:
@pwolfram, I just made a few changes, including your request from above.

I now make sure that "features" come last in a feature collection and "geometry" comes last in a feature.

xylar:
flake8 seems to think the formatting is okay.

xylar:
Off to bed...

pwolfram:
@xylar, I made one small modification at https://github.com/pwolfram/geometric_features/tree/sort_geojson_keys that you can take or leave-- it basically just sorts properties in a more intuitive way setting the order of some properties.  Otherwise, I think this is good to merge.  If you like the change in commit `7fe4eedeb98111b07f0b7394c384ed676cc5395a`, please feel free to cherry-pick and squash.  This is great and looks good to me.  Thanks @xylar!

xylar:
@pwolfram, yes, that makes sense.

xylar:
We will need to redo #71 to be consistent with these changes.  I will do this as soon as this PR is merged.

pwolfram:
Thanks @xylar!

PR #79, opened 2016-11-03
Remove small polygons from MOC basins

xylar:
With this merge, setup_ocean_basins removes small polygons
from MOC basins that are not useful for MOC computations.


xylar:
@milenaveneziani, this PR is just the parts of #78 that seem to clearly be useful.  The disconnected bits of the Tasman Sea, that Amazon River, etc. are removed from the MOC basins so they are made up of only 1 (Atlantic, Indian, and 65N to 65S) or 2 (Pacific and Indo-pacific because of crossing 180 degrees longitude) large polygons.

As we discussed in #78, we will explore another approach to finding the southern boundary that doesn't involve geometric features.

This PR should be ready to test and merge.

milenaveneziani:
New MOCBasins look good to me: https://gist.github.com/milenaveneziani/841c216e44d24c5000752d18952cfd07

Issue #63, opened 2016-09-29
Potentially remove fix_features_at_prime_meridian.py


pwolfram:
Following discussions with @xylar, `fix_features_at_prime_meridian.py` is not used in current scripts and may not be needed anymore.  If this is the case, we should consider removing this script (or merging it with `fix_features_at_antimeridian.py`).

xylar:
Addressed by #82

Issue #57, opened 2016-09-29
Update README following refractor of scripts


pwolfram:
Following addition of new features, workflows, etc.  The README should be updated to better reflect the capabilities of the repository.

cc @xylar

xylar:
Addressed by #82

PR #82, opened 2017-02-13
PEP8 cleanup and update to README.md

xylar:
Clean up PEP8 formatting of each script, driver script and utility file.

Update the README.md file to document (minimally) each script.

Remove the unused fix_features_at_prime_meridian.py script.


xylar:
This PR is intended to address #54, #57 and #63.

xylar:
### Testing

I ran all the manipulation script and verified (by opening or plotting each geojson file) that they produced the expected results.  Given that only formatting was being tested, the important thing is that they ran without errors.

xylar:
@pwolfram, this should be an easy merge.  I think all you need to do is verify that you're okay with the update to the README (or suggest an alternative change) and make sure the basic workflow (merge, plot, split) works for you.

There's no rush on this one so get to it when you have a free moment.

xylar:
@milenaveneziani, given that @pwolfram is pretty busy right now and that these changes are just cosmetic, would you be willing to review (and perhaps merge) this PR instead?

One check would be to make sure all the scripts in the main directory and in util are, in fact, PEP8 compliant.

I could also make the scripts in driver_scripts compliant if you would like.

xylar:
I went ahead and did PEP8 (and a bit of other) clean up on the driver scripts.  I made sure all 3 that I modified run correctly.

milenaveneziani:
@xylar: this looks good to me. I ran flake8 on all scripts, and the only error I got was a:
`E402 module level import not at top of file`
on plot_features for line 26-29, but I don't understand the reason (the module imports *are* at the top of the file..)

xylar:
@milenaveneziani, yeah, there's nothing we can do about that.  It's complaining because, in the following lines:
```
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
```
We call a function (`matplotlib.use`) before an import.  But that's how `matplotlib` backends work.  There's no feasible way around it.

xylar:
Thanks again, @milenaveneziani!

Issue #83, opened 2017-02-13
Region file corresponding to hard-wired regions


q-rai:
Hi!

As part of my work on adding and testing region file support to some analysis members, I've created geojson files for the hard-wired regions that are used for some analysis members (e.g. Water Mass Census, and Surface Area Weighted Averages).

I propose to add them to this repository, I'm just not sure where exactly and under which name.

Currently, this is a region group containing:
- Arctic: lat >= 60
- Equatorial: -15 <= lat <= 15
- Southern Ocean: lat <= -50
- Nino 3: -5 <= lat <= 5; 210 <= lon <= 270
- Nino 4: -5 <= lat <= 5; 160 <= lon <= 210
- Nino 3.4: -5 <= lat <= 5; 190 <= lon <= 240
- Global

[This is what they look like](https://gist.github.com/q-rai/59ef3e4310e97f8125c14112d7a4e41f)

I can:
- Add the entire region group (I propose to put that in `ocean/region` as `Hard-Wired_Regions.geojson`); **or**
- Add each region individually, as it's already the case for other regions.

.. or something entirely different; I'll do whatever would be most useful.

Let me know what you prefer!

xylar:
Hi @q-rai, the right way to do this would be the following:
1. put your `OceanRegions.geojson` file in the root of a new branch of `geometric_features`.
2. rename each features to have a unique, descriptive suffix.  For example, your Arctic region needs to be distinct from the `Arctic Ocean` region we already have, and the same for the `Southern Ocean`.  I don't know what the right suffix might be.  It needs to be more descriptive of what it will be used for than, for example, 'Region' or 'Hard-Wired'.  I thought of 'Analysis' but I'm not sure it's clear what the distinction between `Southern Ocean` from `Southern Ocean Analysis` would be...
3. Give the same tag to all the features (something not used for other features).  This will allow you to regroup the features into a single feature file later on.  You can do this with:
`./tag_features.py -f OceanRegions.geojson -t <tag> -o TaggedOceanRegions.geojson`
The tag will presumably be something similar to the suffix you add to all the feature names in 1.
4. Run `./split_features.py -f TaggedOceanRegions.geojson` to put the features into their proper place.
5. use `git add`, `git commit`, etc. to add all the split features to the repo as normal.

Later, when you wont to recreate your single file with all regions, you would do something like:
```
./merge_features.py -d ocean/region/ -t <tag> -o OceanRegions.geojson
[...]

PR #84, opened 2017-02-15
Add 4 new regions (Equatorial and 3 Nino)

xylar:
Rename setup_ocean_basins to setup_ocean_region_groups and
add these 4 new regions to the groups generated by this script.


xylar:
@q-rai, @mark-petersen and @milenaveneziani, could you take a look at this and make sure you think it addresses the needs of #83.

driver_scripts/setup_ocean_region_groups.py
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env python
 
 """
-This script creates 6 major ocean basins and 5 regions used for computing the
-meridional overturning circulation (MOC).  No arguments are required. The
-optional --plot flag can be used to produce plots of each basin or region.
+This script creates 7 major ocean basins, 5 regions used for computing the
```
milenaveneziani: I would add more detail here: "This script creates the following ocean region groups: i) OceanBasinRegionsGroup, which includes the Atlantic, Pacific, Indian, Arctic, Southern Ocean, Equatorial (global 15S-15N), and Mediterranean basins; ii) MOCBasinRegionGroup, which includes five regions used for computing the meridional overturning circulation (MOC) and meridional heat transport (MHT); and iii) NinoRegionGroups, which includes the Nino3, Nino4, and Nino3.4 regions."

q-rai:
As discussed in #83,  layerWeightedAvgs, surfaceAreaWeightedAvgs an waterMassCensus will need a global region, whereas MHT and MOC don't. I suggest to add a global region to this.

driver_scripts/setup_ocean_region_groups.py
```diff
@@ -91,7 +90,7 @@ def remove_small_polygons(inFileName, outFileName, minArea):  # {{{
     write_all_features(features, outFileName, indent=4)  # }}}
 
 
-def build_basins():  # {{{
+def build_ocean_basins():  # {{{
     basinGroupName = 'OceanBasinRegionsGroup'
     basinFileName = 'oceanBasins.geojson'
```
milenaveneziani: We also agreed it would be good to add the GlobalOcean to this group.

xylar: @milenaveneziani, should this feature just be -180 to 180 lon, -90 to 90 lat or should it be the union of all ocean basins without land?

xylar: I'm going to assume the latter (just the union of all basins) which is slightly simpler.

xylar:
Okay, I've updated the docstring as requested and added a 'Global Ocean' feature to the ocean basins.

driver_scripts/setup_ocean_region_groups.py
milenaveneziani: @xylar: isn't the end result the same?

driver_scripts/setup_ocean_region_groups.py
```diff
@@ -1,12 +1,17 @@
 #!/usr/bin/env python
 
 """
-This script creates 6 major ocean basins and 5 regions used for computing the
-meridional overturning circulation (MOC).  No arguments are required. The
-optional --plot flag can be used to produce plots of each basin or region.
+This script creates the following ocean region groups:
+i) OceanBasinRegionsGroup, which includes the Atlantic, Pacific, Indian,
+Arctic, Southern Ocean, Equatorial (global 15S-15N), and Mediterranean basins;
```
milenaveneziani: sorry, need to add "the Global Ocean"

driver_scripts/setup_ocean_region_groups.py
q-rai: @xylar In the hard-wired regions it's implemented as 'take all cells', which I assume means the former. Does MPAS-O do anything for land cells or are they automatically omitted?

xylar: The only case in which this would matter is if there are cells that are not land in MPAS-Ocean but also don't fall into any of the basins defined in this repo.  This can happen because the definition of the land currently doesn't exactly match up to the definition of the ocean basins, leaving little gaps in some places.  So the question is, do you want to be sure that the global ocean is really the sum of the regional basins (excluding the overlapping Equatorial Basin) or would you rather be sure that every single cell is included.

> In the hard-wired regions it's implemented as 'take all cells', which I assume means the former. 

Yes, I'm sure the hard-wired regions would be more consistent with the latter, but I don't think we want to use the hard-wired regions as a guide because I think they were more like temporary hacks that were easy to add directly in the code.

driver_scripts/setup_ocean_region_groups.py
xylar: Added.

driver_scripts/setup_ocean_region_groups.py
milenaveneziani: I think it is more important that GlobalOcean really is global, more than the sum of all the others.
I will check that during testing before merging.

xylar: Okay, then we need to add another feature, perhaps `Global Ocean 90S to 90N`? Or just `Global Ocean`?

xylar: Okay, I went ahead and switched to a `Global Ocean` feature, consistent with the hard-wired feature.

milenaveneziani:
Thanks @xylar.
I wanted to visualize the oceanBasin feature using a gist, but I get an error (maybe because it's too big a file?).
So I decided to rerun the driver with the --plot option, but I get this error:
`plot type: cyl
 : cannot connect to X server edison12:57.0
  File "./driver_scripts/setup_ocean_region_groups.py", line 272, in <module>
    build_ocean_basins()
  File "./driver_scripts/setup_ocean_region_groups.py", line 135, in build_ocean_basins
    '-o', '%s_Basin.png' % oceanName, '-m', 'cyl'])
  File "./driver_scripts/setup_ocean_region_groups.py", line 36, in spcall
    return subprocess.check_call(args, env=os.environ.copy())  # }}}
  File "/global/project/projectdirs/acme/software/anaconda_envs/edison/lib/python2.7/subprocess.py", line 186, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['./plot_features.py', '-f', 'temp_combined_basins.geojson', '-o', 'Atlantic_Basin.png', '-m', 'cyl']' returned non-zero exit status 1`

Do you understand what the problem may be?

I am running this on edison, and I do have the cartopy package in my python environ.

xylar:
I don't know what the error is for sure but it sounds like it's trying to use xwindows for some reason.

You can visualize using geojson.io.  I think that handles even very large files.

xylar:
Yeah, the error is that you're not using x-forwarding.  If you want to avoid that, you would need to add these lines to plot_features before `import matplotlib.pyplot as plt`:
```
import matplotlib
matplotlib.use('Agg')
```

It might be worth posting this as an issue, because we should support plotting without x-forwarding.

milenaveneziani:
Yes, that did the trick @xylar.
I will submit a quick PR asap.

Merging this one now, as all looks good.

PR #85, opened 2017-02-15
Allow plot_features to produce png files without x-forwarding

milenaveneziani:
Tiny change here to allow plot_features to work without x-forwarding.


xylar:
I believe the original thinking was that we would support plotting to the screen as well as to png files.  But I don't see any evidence that this is possible now.  So I see no harm in making sure we always use the `Agg` backend.

PR #86, opened 2017-02-16
Add 'seed_point' tag to all ocean points

xylar:
This tag is used to find seed points for the "flood fill" operation
used to make sure the ocean is a single, connected body when setting
up MPAS-Ocean `global_ocean` test cases.

Adding this tag will make it possible to add ocean points that are
not necessarily in the ocean domain for all model configurations.
For example, points below ice shelves would not be in the ocean
domain for setups without ice-shelf cavities.

### DIFF
--- a/ocean/point/Equatorial_Pacific_E145.0_N00.0/point.geojson
+++ b/ocean/point/Equatorial_Pacific_E145.0_N00.0/point.geojson
@@ -5,7 +5,7 @@
             "type": "Feature", 
             "properties": {
                 "name": "Equatorial_Pacific_E145.0_N00.0", 
-                "tags": "", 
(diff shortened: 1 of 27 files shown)


xylar:
This change is needed to help address https://github.com/MPAS-Dev/MPAS/issues/1247, allowing points to be added to `ocean/point` without the restriction that they are in the ocean for all possible `global_ocean` configurations of MPS-Ocean.

xylar:
@milenaveneziani, this should be a very quick review and merge (I hope!).

milenaveneziani:
There is really nothing much to review here :)

xylar:
Thanks @milenaveneziani!

PR #89, opened 2017-05-02
Moved Global_65S_65N from MOCBasins to oceanBasins

milenaveneziani:
This is to fix a small bug in setup_ocean_region_groups where the Global ocean region Global_65S_65N was included in the MOCBasinsGroup rather than in the oceanBasinsGroup.


xylar:
Yep, makes sense to me.

milenaveneziani:
Thanks you. Yes, I hadn't realized that that global region (which we introduced for the OHC calculation, since there are observations for the global OHC between 65S and 65N) ended up in the wrong group.

PR #90, opened 2017-08-25
Add driver script for merging ice-shelf features

xylar:
This driver merges together regions to make up 100 individual ice shelves (some of which are divided into multiple region files because they cross several IMBIE basins) and 6 analysis regions (currently: Antarctica, Peninsula, West Antarctica, East Antarctica, Filchner-Ronne and Ross).

The resulting file can be used to create masks for computing ice-shelf time series for individual ice shelves.

### DIFF
--- a/driver_scripts/setup_ice_shelves.py
+++ b/driver_scripts/setup_ice_shelves.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python
+
+"""
+This script creates region groups for ice shelves
+
+Author: Xylar Asay-Davis
+"""
+
+import os
+import os.path
+import subprocess
+
+
+def build_ice_shelves():  # {{{
+    groupName = 'IceShelvesRegionsGroup'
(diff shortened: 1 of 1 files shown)


driver_scripts/setup_ice_shelves.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
+
+"""
+This script creates region groups for ice shelves
+
```
stephenprice: I don't know if it is worth noting in the comments here that this script needs to be run from the root level directory of 'geometric_features'? Perhaps that is already implied by all of the scripts that sit in the 'driver_scripts' subdir?

stephenprice:
@xylar -- I believe this is working as advertised (see my plot below of the resulting feature file w/ all ice shelves). There's one minor comment I made that is not critical right now, so I will go ahead and merge this in a bit.

temp

PR #91, opened 2017-09-14
Add standard transect sections

xylar:
Also adds a script that was originally used to create the sections (for provenance) and a driver script for merging the features together into a region group based on their common tag.

### DIFF
--- a/driver_scripts/setup_ocean_standard_transport_sections.py
+++ b/driver_scripts/setup_ocean_standard_transport_sections.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+
+"""
+This script creates region groups for the standard set of transport sections,
+all of which have the "standard_transport_sections" tag.
+
+Author: Xylar Asay-Davis
+"""
+
+import os
+import os.path
+import subprocess
+
+tag = 'standard_transport_sections'
+fileName = 'standard_transport_sections.geojson'
+groupName = 'StandardTransportSectionsRegionsGroup'
+    
+if os.path.exists(fileName):
+    os.remove(fileName)
+
+
+subprocess.check_call(['./merge_features.py',
+                       '-d', 'ocean/transect',
+                       '-t', tag,
+                       '-o', fileName])
+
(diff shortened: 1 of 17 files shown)


xylar:
This PR is a step toward addressing the following issue in MPAS-Analysis: https://github.com/MPAS-Dev/MPAS-Analysis/issues/242

xylar:
@mark-petersen, I didn't like the very long and poorly formatted names of the transects.  Instead, I included the extra information as a "note" within each feature.  Please let me know if you would prefer an alternative solution.

xylar:
@mark-petersen, just pinging you on this PR, in case you have time for it in the next few days.  It's not terribly urgent but should be a quick review.

xylar:
## Testing

I ran:
```bash
./driver_scripts/setup_ocean_standard_transport_sections.py
```

The resulting geojson file can be plotted on geojson.io as follows:

image

Then, I used the mask creator (in the MPAS-Tools repository) with a standard oQU240v3 mesh to create a mask NetCDF file (taken from E3SM's inputdata repository):
```bash
./MpasMaskCreator.x oQU240v3_mesh.nc oQU240v3_standardTransectMasks.nc -f standard_transport_sections.geojson
```
To examine the transects, I used the paraview extractor tool:
```bash
./paraview_vtk_field_extractor/paraview_vtk_field_extractor.py -m oQU240v3_mesh.nc \
    -f oQU240v3_standardTransectMasks.nc -v all -d maxEdges= TWO= vertexDegree= \
    maxEdges2= nVertLevels=0 nTransects=: -o vtk_standard_transects
```

Here are a few example plots:
image
image


The resulting transects have 2 issues that should be addressed in the MPAs-Tools repo:
* They are zig-zags rather than following the shortest path between the endpoints: https://github.com/MPAS-Dev/MPAS-Tools/issues/143
* The names of the transects are not stored in the NetCDF file: https://github.com/MPAS-Dev/MPAS-Tools/issues/184

These do not need to be addressed before merging this PR.

PR #92, opened 2017-09-29
Add IMBIE basins to ice-shelf driver

xylar:
@JeremyFyke would like to use these for correlation with
atmospheric forcing in each basin.

### DIFF
--- a/driver_scripts/setup_ice_shelves.py
+++ b/driver_scripts/setup_ice_shelves.py
@@ -136,6 +136,11 @@ def build_ice_shelves():  # {{{
                           'West Antarctica': ['WestAntarcticaIMBIE'],
                           'East Antarctica': ['EastAntarcticaIMBIE']}
 
+    nIMBIEBasins = 27
+    for basinNumber in range(1, nIMBIEBasins+1):
+        basinName = 'Antarctica_IMBIE{}'.format(basinNumber)
+        combinedIceShelves['IMBIE{}'.format(basinNumber)] = [basinName]
+
     # build analysis regions from combining ice shelves from regions with the
     # appropriate tags
     for shelfName in combinedIceShelves:


xylar:
@JeremyFyke, if you could please run 
```
./driver_scripts/setup_ice_shelves.py
```
and verify that it runs for you and that the resulting file contains the IMBIE basins, that would be great!  You can either use the plotting script in the repo or upload `iceShelves.geojson` to http://geojson.io and plot it there.

Note, you may need to conda install any libraries you're missing that this repo uses.

If you approve, I'll merge this myself.

JeremyFyke:
Hey Xylar, tool runs successfully.  Do you know if there a way to extract just the IMBIE basins for plotting on geojson.io?  I'm finding it a bit difficult to visualize these via web or provided Python script (not sure this script set up to visualize these new basin designations yet, but I could be wrong here).

xylar:
@JeremyFyke, there's not a clean and easy way to do what you want, but this should do it:
```bash
./split_features.py -f iceShelves.geojson -o temp
cd temp/region/
mkdir ../temp
mv IMBIE* ../temp
rm -rf *
mv ../temp/IMBIE* .
cd ../..
./merge_features.py -d temp/region/ -o imbie.geojson
```
This splits out all the features in `iceShelves.geojson` into individual features, deletes all the ones that don't start with IMBIE and then merges them back together.

Here's what I see on http://geojson.io when I do that:
image

JeremyFyke:
Thanks Xylar.  I think this looks reasonable (assuming that the underlying geojson.io continent is messed up, leading to the unequal longitude scaleing on the edges...)

I'd go ahead and merge - then, I'll do more sanity checking during analysis build-out.

jer

xylar:
@JeremyFyke,  I'm not sure what you mean by "unequal longitude scaling on the edges" but the projection is the one Google Maps uses (Mercator) which definitely does strange things as it approaches the poles.  That's partly why we have our own plotting tool in this repo.  With:
```
./plot_features.py -f imbie.geojson -m southpole
```
I get:
imbie
Does that look right to you (e.g. less distorted than geojson.io)?

xylar:
By the way, to approve, please agree to review this PR at the top and choose "approve", possibly leaving a comment.  Thanks!

Issue #87, opened 2017-03-01
plot_features.py does not appear to support points


xylar:
On IC, I am getting an error when trying to plot a features file containing points:
```
$ rm features.geojson
$ ./merge_features.py -d ocean/point/ -o features.geojson
$ ./plot_features.py -f features.geojson -m cyl
plot type: cyl
  feature: Equatorial_Pacific_E145.0_N00.0
  feature: Equatorial_Pacific_E155.0_N00.0
  feature: Equatorial_Pacific_E165.0_N00.0
  feature: Equatorial_Pacific_E175.0_N00.0
  feature: Equatorial_Pacific_W090.0_N00.0
  feature: Equatorial_Pacific_W095.0_N00.0
  feature: Equatorial_Pacific_W105.0_N00.0
  feature: Equatorial_Pacific_W115.0_N00.0
  feature: Equatorial_Pacific_W125.0_N00.0
  feature: Equatorial_Pacific_W135.0_N00.0
  feature: Equatorial_Pacific_W145.0_N00.0
  feature: Equatorial_Pacific_W155.0_N00.0
  feature: Equatorial_Pacific_W155.0_N02.0
  feature: Equatorial_Pacific_W155.0_N04.0
  feature: Equatorial_Pacific_W155.0_N06.0
  feature: Equatorial_Pacific_W155.0_N08.0
  feature: Equatorial_Pacific_W155.0_N10.0
  feature: Equatorial_Pacific_W155.0_S02.0
  feature: Equatorial_Pacific_W155.0_S04.0
  feature: Equatorial_Pacific_W155.0_S06.0
  feature: Equatorial_Pacific_W155.0_S08.0
  feature: Equatorial_Pacific_W155.0_S10.0
  feature: Equatorial_Pacific_W165.0_N00.0
  feature: Equatorial_Pacific_W175.0_N00.0
  feature: Southern_Ocean_Drake_Passage_W070.0_S55.0
  feature: Southern_Ocean_Drake_Passage_W070.0_S60.0
  feature: Southern_Ocean_Drake_Passage_W070.0_S65.0
saving features.png
Traceback (most recent call last):
  File "./plot_features.py", line 265, in <module>
    plot_features_file(args.features_file, mapInfo, max_length=args.max_length)
  File "./plot_features.py", line 215, in plot_features_file
    fig.savefig(plotFileName)
[...]

Issue #53, opened 2016-09-26
Plotting error with moll projection for combined Atlantic and Southern Ocean


pwolfram:
Using the file at https://gist.github.com/43b8c23a27a84da35eadd9304a1b3753, there is a plotting error for the moll projection (which does not show up for moll2):

![screenshot 2016-09-26 16 18 41](https://cloud.githubusercontent.com/assets/4295853/18853904/fbe37422-8404-11e6-8e99-7a51c6bcd7eb.png)

Script to reproduce error from repository root:

```bash
mkdir test_feature_collection
cd test_feature_collection/
ln -s ../ocean/region/North_Atlantic_Ocean/region.geojson NA_region.geojson
ln -s ../ocean/region/South_Atlantic_Ocean/region.geojson SA_region.geojson
ln -s ../ocean/region/Southern_Ocean/region.geojson SO_region.geojson
cd ..
./merge_features.py -d test_feature_collection/ -o AO_SO.geojson
./plot_features.py -f AO_SO.geojson 
```

Note, found in the process of testing #52.  I believe this is a plotting error but it could be an error with preliminary changes in #52.

pwolfram:
Going further:

```bash
./difference_features.py -f AO_SO.geojson -m landice/region/Antarctica_IMBIE1/region.geojson -o AO_SO_masked.geojson
./plot_features.py -f AO_SO_masked.geojson 
```

moll is even more visibly messed up (geojson file at https://gist.github.com/79c3cc90ef72ad8320e799c4a54eb203)...

![screenshot 2016-09-26 16 39 28](https://cloud.githubusercontent.com/assets/4295853/18854331/d6425f14-8407-11e6-83a1-dd29efedb4d9.png)

xylar:
@pwolfram, I can reproduce these bugs and will look into them when I have a chance.

xylar:
@pwolfram, I strongly suspect this is a bug in how shapes including the south (or presumably also the north) pole are handled in cartopy for this projection.  It should be pretty easy to set up an example where we plot a box similar to the SouthernOcean in a simplified script to indicate the problem.  Note that there are glitches in the Antarctic land coverage in both `moll` and `moll2` projections as well.
Anyway, I don't think this is an issue for this repo.  Please close if you agree and open an issue with cartopy instead.

xylar:
@pwolfram, here is a simpler script demonstrating that the problem is with cartopy, not with our plotting routine:
[...]

PR #94, opened 2017-10-06
Fix plotting of points

xylar:
Also, adjust standard figure size, dpi and some plotting properties for transects

### DIFF
--- a/plot_features.py
+++ b/plot_features.py
@@ -57,7 +57,6 @@ def build_projections():  # {{{
 
     return projections  # }}}
 
-
 def plot_base(mapType, projection):  # {{{
 
     ax = plt.axes(projection=projection)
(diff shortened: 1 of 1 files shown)


xylar:
This PR is intended to address #87

xylar:
## Testing

```bash
./merge_features.py -d ocean/point/ -o points.geojson
./plot_features.py -f points.geojson
./driver_scripts/setup_ocean_standard_transport_sections.py 
./plot_features.py -f standard_transport_sections.geojson
```
(Examples here shortly...)

xylar:
I just realized things aren't working for many of the point plots.  I'll keep working on this...

pwolfram:
Verified that points, transects, and regions worked.

```
./merge_features.py -d ocean/region/Southern_Ocean/ -o SO.geojson
./plot_features.py -f SO.geojson 
```

Note, strange behavior with the moll projection because there is no green polygon-- @xylar, does a reason why this doesn't plot correctly immediately come to mind?  We may have a bug here because moll2 works.  Is so, I can open an issue.
so_moll2

so_moll

xylar:
Thanks @pwolfram.  I'm not concerned about the moll plot.  There's something glitchy in cartopy we had a long discussion about it awhile ago.  But I just don't think we really need this particular projection to work.  One option would be to take it out (moll2 as well).

pwolfram:
Forgot about https://github.com/MPAS-Dev/geometric_features/issues/53 and https://github.com/SciTools/cartopy/issues/802, thanks for reminding me!  I don't think we need to do anything on this right now unless we run the risk of being confusing by having moll and moll2, but this is low priority and not 100% clear to me.

PR #95, opened 2017-10-25
Add English Channel Deepen

mark-petersen:
When creating the V3 mesh, the English Channel was closed on the
EC60to30 mesh. I drew a line by hand through the middle of the English
Channel. This version was used to make the current EC60to30V3 mesh, so
really needs to be in this geometric features repo with these points.

Note that the middle commit with 15m was used for the meshes made in
spring of 2017.  We changed it to 120m here to be more realistic.

### DIFF
--- a/ocean/transect/English_Channel_Deepen/transect.geojson
+++ b/ocean/transect/English_Channel_Deepen/transect.geojson
@@ -1,0 +1,29 @@
+{
+    "type": "FeatureCollection", 
+    "features": [
+        {
+            "type": "Feature", 
+            "properties": {
+                "name": "Red Sea Deepen", 
+                "tags": "Red_Sea_Deepen;Critical_Passage", 
+                "object": "transect", 
+                "component": "ocean", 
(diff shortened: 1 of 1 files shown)


xylar:
Wikipedia says the average depth at it's widest point is 120 m.  That seems like as good a number as any.  The deepest point is 180 m deep while other parts are quite a bit shallower (45 m) so 120 m would be a reasonable compromise.

xylar:
As we discussed, I would suggest showing a plot of the resulting QU240 or QU120 mesh to show that the channel is, indeed, open.  Once that's posted, feel free to merge.

mark-petersen:
Mesh without this new channel deepening on the EC60to30:
screen shot 2017-10-25 at 4 29 17 pm

Mesh with this PR in channel deepening:
screen shot 2017-10-25 at 4 29 54 pm

I am preparing and testing all the scripts in an EC60to30 directory.  I will test other resolutions when I've made more progress there.

PR #97, opened 2018-02-27
Add feature for all IMBIE basins and ice shelves

stephenprice:
Add a new combined feature file for all Ant. ice sheet drainage basins
and their associated ice shelf extents, as defined by IMBIE. This was
done using merge_feature.py to 1) extract all IMBIE drainage basin
regions (by tag) and then 2) combine them into a single new region.
It has been used to define regional masks for a 20 km Ant. ice sheet
mesh and tested using those masks along with with the MPAS-LI regional
stats analysis member.


pwolfram:
Looks good to me, there was no problem generating the figure.
region_southpole

xylar:
I’m confused about the purpose here. Do you want a single feature that has the total size of antarctica out to the continental shelf. If you truly had combined features, that is, fused the geometry into a single polygon, that’s what you would get. But I’m unclear on why such a feature would be useful (except perhaps occasionally).

If you didn’t fuse the basins and just merged them, I really don’t think we want to include that in the repo. Instead, please make a driver script to do this.

stephenprice:
@xylar -- Thanks for the input. This is exactly the guidance I was looking for as I'd forgotten the philosphy for adding functionality to this repo. I can close this PR and submit a PR for the relevant script to build the feature file instead.

To be clear, what we need is what Philip has plotted above -- a single .geojson file that contains all of the indiv. basin / ice shelf information, which can then be used w/ MPAS-LI to generate regional stats over those basins / ice shelves (which I've already confirmed works in a realistic application).

PR #98, opened 2018-02-27
WIP: add extended shelf option to Ant. basins driver

stephenprice:
Add an optional argument, 'extended_shelves', to existing
setup_antarctic_basins.py driver script so that extended ice shelf areas
in iceshelves regions files are included in basin region definitions.

### DIFF
--- a/driver_scripts/setup_extended_antarctic_basins.py
+++ b/driver_scripts/setup_extended_antarctic_basins.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+"""
+This script combines Antarctic basins into a single feature file.  No arguments
+are required. The optional --plot flag can be used to produce plots of the
+result.
+"""
+import os
+import os.path
+import subprocess
+from optparse import OptionParser
+
+
+parser = OptionParser()
+parser.add_option("--plot", action="store_true", dest="plot")
+
+options, args = parser.parse_args()
+
+outFinalFileName = 'Extended_Antarctic_Basins.geojson'
+
+if os.path.exists(outFinalFileName):
+    os.remove(outFinalFileName)
+
+tempFileName = 'merged_basin.geojson'
+if os.path.exists(tempFileName):
+    os.remove(tempFileName)
+
+for basin in range(1, 28):
+    print "Adding feature from IMBIE basin {:d}".format(basin)
+    outFileName = 'AntarcticIMBIE{:d}.geojson'.format(basin)
+    basinName = 'Antarctica_IMBIE{:d}'.format(basin)
+    tags = basinName
+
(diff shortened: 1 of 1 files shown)


stephenprice:
@xylar -- Your existing setup_antarctic_basins.py script does 90% of what I'm looking for here so I thought it would make sense to just add some functionality to that. It should be self explanatory based on the description in the commit message. But to reiterate, I want to add the option to include your _extended_ ice shelf regions in the basin area definitions.

However, I'm having problems getting it to work. Pulling the IMBIE basin info from the regions in the iceshelves/ dir requires some different syntax relative to what is in the script already, and I'm trying to use the 'tag' option of merge_features.py for that. But, I've never gotten this option to work for other than a single tag (a list of multiple tags doesn't seem to work). 

You can sort of see that here w/ my debug lines. If you include (for example) only 'Antarctica_IMBIE23' in the tag list, you get basin 23 out in the final features file. However, if you include 'Antarctica_IMBIE23;Antarctica_IMBIE22' in the tag list, you get nothing but a file header. I've tried various other syntax here too (e.g. double quotes) but the tag option still doesn't seem to work for multiple items. Maybe there's something obvious wrong with my syntax though? Copying @pwolfram in case he can comment while you are away. Thanks!

xylar:
@stephenprice, at the risk of being a jerk who tells you, "there's a better way to do this and I've gone ahead and redone it for you", there's a better way to do this and I've gone ahead and redone it for you:
https://github.com/xylar/geometric_features/blob/add_extended_shelf_option_to_setup_antarctic_basins_driver/driver_scripts/setup_extended_antarctic_basins.py

A couple of comments:
* The way tags work in `merge_features.py`, I think you can specify a directory and a list of tags but if you specify files it just uses the files and ignores the tags.  That is likely something we should change...
* I think based on what you've described, it would be necessary to 1) merge feature for a given basin; 2) combine those features into a single feature with the name of the basin and 3) merge all of the basins into a single file. If I've got that right, that's what my script does.
* The behavior of the new script is different enough from `setup_antarctic_basins.py` that I think it makes sense for this to just be a different script.

Feel free to do a `git reset --hard xylar/geometric_features/dd_extended_shelf_option_to_setup_antarctic_basins_driver` if you're happy with my version.  Alternatively, feel free to use mine as a starting point for rewriting your own.

Here's the result of the script:
extended_antarctic_basins

stephenprice:
@xylar -- I don't need to reinvent the wheel here. This looks fine to me. I don't know if you want to do a PR for this or just merge it on your own. Either is fine w/ me. You can tag me if you want someone else to merge it. I've already tested it and confirmed it works as needed. Thanks.

PR #99, opened 2018-05-28
Fix incorrect file naming in land coverage driver

xylar:
A clean-up pull request did an incomplete job of renaming the output file, resulting in an incorrect land coverage feature file.

### DIFF
--- a/driver_scripts/setup_ocean_land_coverage.py
+++ b/driver_scripts/setup_ocean_land_coverage.py
@@ -18,15 +18,15 @@ parser.add_option("--with_cavities", action="store_true", dest="with_cavities")
 
 options, args = parser.parse_args()
 
-defaultFileName = 'features.geojson'
-outName = 'landCoverage.geojson'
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen, I noticed this when working on something else but I think this might affect the pull requests you asked me to review (https://github.com/MPAS-Dev/MPAS-Tools/pull/220 and https://github.com/MPAS-Dev/MPAS-Model/pull/49).

xylar:
## Testing

Before this fix, two separate files, `features.geojson` and `landCoverage.geojson` are produced, the first containing the coastline north of 60S and the second containing only the coastline below 60S.  After thie fix, the global coastline is contained in `landCoverage.geojson` and  `features.geojson` is not produced.

landcoveragewithoutcavities

mark-petersen:
Now only produces the single files landCoverage.geojson:
image

xylar:
Thanks @mark-petersen!

PR #103, opened 2019-02-27
Add the two new critical passages and one "critical land blockage"

xylar:
These features are needed for the CONUS grid in E3SM and perhaps for future versions of our EC60to30 and similar lower resolution meshes.

### DIFF
--- a/ocean/transect/Antarctic_Peninsula/transect.geojson
+++ b/ocean/transect/Antarctic_Peninsula/transect.geojson
@@ -1,0 +1,61 @@
+{
+    "type": "FeatureCollection", 
+    "features": [
+        {
+            "type": "Feature", 
+            "properties": {
+                "name": "Antarctic Peninsula", 
+                "tags": "Antarctic_Peninsula;Critical_Land_Blockage", 
+                "object": "transect", 
+                "component": "ocean", 
+                "author": "Xylar", 
+                "height": "100.0"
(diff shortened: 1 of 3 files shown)


xylar:
You can view the files:
https://github.com/MPAS-Dev/geometric_features/blob/60af68550eb3b69df702d6246c9a81d2f4fcec94/ocean/transect/Antarctic_Peninsula/transect.geojson

https://github.com/MPAS-Dev/geometric_features/blob/60af68550eb3b69df702d6246c9a81d2f4fcec94/ocean/transect/Nares_Strait/transect.geojson

https://github.com/MPAS-Dev/geometric_features/blob/60af68550eb3b69df702d6246c9a81d2f4fcec94/ocean/transect/White_Sea/transect.geojson

xylar:
Andrew approved the depths (250 for Nares, 50 for White) but I'm open to changes.

xylar:
It looks like it will be necessary to have the same `depth` property in all critical passages and critical land blockages in order for the mask tool to handle them in the expected way.  (It would be possible to process critical land blockages completely separately but I think it's much better if they are handled along with critical passages but with a positive depth, rather than a negative one).

We could rename the depth property to something else but I think it's probably fine as it is.  

**Note**: this will break backwards compatibility with compass, so that's a consideration with these changes.

PR #104, opened 2019-05-08
Add Antarctic ocean regions

xylar:
These are from [Timmermann and Hellmer (2013)](http://link.springer.com/10.1007/s10236-013-0642-0) and will be used to create plots in MPAS-Analysis like those in the same paper, addressing https://github.com/MPAS-Dev/MPAS-Analysis/issues/571


xylar:
Features can be reviewed by looking here:
https://github.com/MPAS-Dev/geometric_features/blob/848ad9091d947459c98c456ecb9c1ab03b3eed34/geometric_data/ocean/region/Amundsen_Sea_Deep/region.geojson

xylar:
> The only one I am worried about is the Weddell shelf, 'cause it seems awfully close to land (in some cases the line is actually above land). 

There are really places where ice shelves stick out past the continental shelf break in this region.  That's what you mean by "above land" I guess?  That's the reality.  I'm following the definition given in Timmermann and Hellmer, and I think we want to stick with that rather than defining our own region.  

> Would we have enough grid points at low-res in those areas?

I agree that there may not be a huge number of points from the eastern Weddell contributing.  But there should be some in front of FRIS at least.

milenaveneziani:
> There are really places where ice shelves stick out past the continental shelf break in this region

yes, you are right, I think it was Brunt I was thinking about.

Maybe we can apply this to the LR and see how we do in the eastern weddell. We could always adjust it (just for the Weddell) from the 800 to the 850 m isobath, for example, if it makes a difference.

xylar:
Okay @matthewhoffman and @milenaveneziani, this is much cleaner now and ready for you to take a look (another look in @milenaveneziani's case) if you like.  It might make sense to not worry too much until I get the corresponding analysis coded up.

@milenaveneziani, you're totally right that we might want to tweak these masks based on what we find in the analysis.  I suspect it wont' be a big problem if there's just not much data from the Eastern Weddell contributing to the Weddell Shelf averages.  I think that was probably the case in Timmermann and Hellmer (2013) as well, and we'd like to be able to make direct comparisons to the degree possible.  But we shouldn't feel too tied to their definitions if we come up with something more useful.

matthewhoffman:
@xylar , this is great! Thanks for putting this together.  I think this will be a useful capability for the Cryosphere campaign in general, and it will be nice to have masks that can be used off the shelf to add new ad hoc analysis outside of MPAS-Analysis.

One thought - it might be useful to add a continental shelf region for EAIS.  I realize it's really narrow and may be of limited use (at low-res) but it would complete the set.

One other thought - would the analysis somehow automatically exclude the subshelf cavities?  Because right now those region edges cut through the middle of FRIS and Ross - I would think we'd want the region to stop at (or approximately at) the calving front as the water properties often change substantially under the shelves.

milenaveneziani:
> would the analysis somehow automatically exclude the subshelf cavities? 

we do have the subshelf cavity regions already, right? So, what you are saying is the the shelf regions should not intersect/overlap with the subshelf cavity regions?

xylar:
> One thought - it might be useful to add a continental shelf region for EAIS. I realize it's really narrow and may be of limited use (at low-res) but it would complete the set.

That would be very easy to add.  I might break it into 2 segments because it's such a large area.  I'll see if any other papers might suggest where the break should be.

> One other thought - would the analysis somehow automatically exclude the subshelf cavities? Because right now those region edges cut through the middle of FRIS and Ross - I would think we'd want the region to stop at (or approximately at) the calving front as the water properties often change substantially under the shelves.

Yes, we would only use points where `landIceMask == 0`.  There would also be some vertical averaging that excludes layers above a certain depth, as in Timmermann and Hellmer (2013).

> we do have the subshelf cavity regions already, right? So, what you are saying is the the shelf regions should not intersect/overlap with the subshelf cavity regions?

The cavity regions actually extend far into the Southern Ocean, we just include only places where `landIceMask == 1`.  These new masks would be the opposite.  But I don't want to make assumptions here about where the calving front/grounding line are because these vary a little bit with resolution and could be different if we used a different data set.

xylar:
Okay, at @matthewhoffman's suggestion I added regions covering East Antarctica:
antarctic_southpole
There are Eastern Weddell and Western Weddell Seas that overlap with the Weddell Sea region above, but I think that break will be useful.  The Eastern Weddell extends to 45 E, covering all the Dronning Maud Land ice shelves.  Then, there is an "East Antarctic Seas" region that covers the remainder around to the Eastern Ross Sea.  Let me know what you think.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
```diff
@@ -0,0 +1,280 @@
...
+    bedmap2_bin_to_netcdf('bedmap2.nc')
+
+    fcContour700 = get_longest_contour(contourValue=-700., author=author)
+    fcContour800 = get_longest_contour(contourValue=-800., author=author)
+
+    fc = FeatureCollection()
+
+    fc.merge(split_rectangle(
+        lon0=-63., lon1=0., lat0=-80., lat1=-65., name='Weddell Sea',
+        author=author, tags=timTags, fcContour=fcContour800))
```
milenaveneziani: @xylar: looking more closely at the region boundaries. I think the full Weddell Sea should extent more to the north and east, because we are cutting through the Weddell Gyre at 65S and not considering the extent east of 0. I think part of the gyre extent to at least 58S, and then to 20 or 30E. I can check more precise numbers before we settle.

milenaveneziani: This would be a different region than the Weddell region considered in the Timmerman paper.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
```diff
@@ -0,0 +1,280 @@
...
+        lon0=160., lon1=180., lat0=-80., lat1=-67., name='Western Ross Sea',
+        author=author, tags=timTags, fcContour=fcContour700))
+
+    fc.merge(split_rectangle(
+        lon0=45., lon1=160., lat0=-80., lat1=-62., name='East Antarctic Seas',
+        author=author, tags=kusTags, fcContour=fcContour800))
+
+    fc.merge(make_rectangle(
+        lon0=-180., lon1=180., lat0=-80., lat1=-60., name='Southern Ocean 60S',
+        author=author, tags=timTags))
```
milenaveneziani: we probably don't need this region (SO 60S), right? it's probably here because that's what wikipedia says the SO is, or something like that, but we always agreed that this did not represent the SO.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
xylar: Yeah, I'm somewhat hesitant to deviate from their definitions.  I could make the definitions of the Western and Eastern Weddell Seas whatever we think is best (since those I'm making up as I go along).  They already extend further east.  I can move the northern boundary to 58 S.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
xylar: Again, this is the definition used by Timmermann and Hellmer.  I'd really like to keep it for comparison with them.

xylar: I could add another plot with the "full" Southern Ocean (which we have extending to 48 S).

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
milenaveneziani: oh I see, I didn't realize that the Weddell region came from the paper. I can check the Western and Eastern Weddell boundaries then, no problem, although probably not right now, if that's OK with you. Also, could we also merge the 2 regions and have a Weddell Gyre region or something like that?

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
milenaveneziani: no, no need to do that. I thought this was a legacy of the first SO region we had originally. I am surprised the paper uses this as SO definition though..

xylar:
@milenaveneziani, here are the regions (not including Weddell Sea Shelf/Deep and Southern Ocean 60S) 
temp_southpole

If you're concerned about me adding a Weddell Sea feature with unrealistic bounds, I could just add the Shelf/Deep versions of Weddell Sea and no the feature that is the sum of those two.  Would that be better?  Or I could replace it with the sum of Eastern and Western Weddell.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
xylar: Done.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
xylar: I expanded the Eastern and Western Weddell further north.

feature_creation_scripts/antarctic_ocean_regions/antarctic_ocean_regions.py
xylar: I went ahead and added zmin/zmax to the full Southern Ocean feature so we can use that too.

xylar:
Thanks @matthewhoffman.  I went ahead and redefined Weddell Sea (but not Weddell Shelf and Weddell Deep, which are the ones I actually plan to use for analysis).  But I agree that we can tweak these to our needs later on.

milenaveneziani:
OK, if I criticize something I need to do some work, right? :) 
Here are two figures from two of Alex Orsi's papers, indicating the southernmost location of the ACC and the Weddell Gyre:
Screen Shot 2019-05-10 at 10 00 22 AM

Screen Shot 2019-05-10 at 10 00 03 AM

Based on these, I would define the Earsten Weddell like this: maxlat=60S, maxlon=20W, and the Western Weddell like this: maxlat=55S, lon=20W-25E. And the total Weddell Gyre region as the sum of the two.

What do you think?

xylar:
@milenaveneziani, thanks for looking at this!

I think you mixed up east and west but I'm good with your numbers if that's true.

milenaveneziani:
indeed I did.. I am one of those people whose brain has problems with identifying left/right, east/west without thinking :D

xylar:
> I am one of those people whose brain has problems with identifying left/right, east/west without thinking :D

Me, too.  Try working around the Ross Sea.  Super confusing!  The Western Ross Sea is closer to East Antarctica.

xylar:
Okay, here is the same plot from above with this updated definition and then the Weddell Sea on its own to confirm that it's the sum of Eastern and Western.

temp_southpole

weddell_southpole

xylar:
Thanks, @milenaveneziani!

> Probably the Ross Gyre will need a similar treatment (extension to the north)

Absolutely true.  Since there's no `Ross Sea` feature currently, we are free to define one later on.

milenaveneziani:
Here are Orsi's papers, for reference:
Orsi et al. 1993, On the circulation and stratification of the Weddell Gyre. DSRI, 40, 169-203.
Orsi et al. 1995, On the meridional extent and fronts of the Antarctic Circumpolar Current. DSRI, 42, 641-673.

xylar:
Thanks, that's good to have.

PR #106, opened 2019-07-30
Added/updated Arctic transects and updated all standard transects

milenaveneziani:
This adds two transects: Davis Strait and Barents Sea Opening. It updates Lancaster Sound transect (with locations we know there is observations to compare with). And it creates an additional Nares Strait transect. The old one has been moved to Nares_Strait_old: note that this transect goes in the 'wrong' direction, west-to-east, instead of going east-to-west, which will produce positive transport for southward rather than northward flow.

Edit: we ended up updating all of the standard transects (hence the updated title of this PR), by  adding several more points on the long ones and also extending the end points into land, so that to avoid open transects in low-resolution meshes.


mark-petersen:
Thanks, @milenaveneziani. I'll test in the next few days with a new mesh.

milenaveneziani:
@mark-petersen, before you do that, I am going through the various steps to merge the transects in one file, plot them (using geometric_features) and then create mask file for a specific mesh (with MPAS-Mask-creator). Once I am done, I will post steps on Confluence, as we agreed yesterday. You can follow the steps and see if they work for you.

milenaveneziani:
ok, first 2 steps were successful. Here is a plot of the merged 4 features: Lancaster (purple), Nares (green), Davis (orange), and Barents Sea Opening (yellow).
ArcticNewTransects_northpole

mark-petersen:
@milenaveneziani, on basin-wide transects, MpasMaskCreator can wander off-course from a great circle or lat/lon line. We should check the actual edge locations for a high-resolution mesh like the Greenland. If needed, we could easily insert some points in between in these geojson files.

milenaveneziani:
Are you worried about a specific transect in particular? Barents Sea?

milenaveneziani:
This is the BSO transect I got for E3SM-Arctic. Didn't think it was too bad:

BarentsSeaOpeningTransect_ARRM60to10

milenaveneziani:
And here is Davis Strait:

DavisStraitTransect_ARRM60to10

milenaveneziani:
I could definitely add some points in there, for these 2.

mark-petersen:
Thanks for the plots! They aren't too bad, but it's so easy to add a few extra points in, that will alleviate any concern about deviating from a straight transect.

milenaveneziani:
Here are the new ones for ARRM60to10. Not sure that they are really 'better' than the first ones, what do you think?

BarentsSeaOpeningTransectNew_ARRM60to10
DavisStraitTransectNew_ARRM60to10

milenaveneziani:
Anyways, we can keep this. I'll push the new ones in a minute.

mark-petersen:
I think it's better, in that across different meshes, it will be closer to the same transect. Thanks!

mark-petersen:
@milenaveneziani (FYI @xylar) I was able to generate transport masks start to finish, with a few hiccups. I tested it on a new EC60to30, created by Jigsaw, that we are using for a FESOM/ICON comparison. (My working name is EC60to30J1.)

I checked out this branch on grizzly and tried this:
```
source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_compass.sh
merge_features -c ocean -b transect  -t "standard_transport_sections" -o standard_transport_sections.geojson 
```
That produces a geojson file tagged with "standard_transport_sections", but it accesses the premade python package, not this branch, so does not include your new ones. I didn't know how to force run a local copy of merge_features (@xylar I don't understand how to call a function locally in `geometric_features/__main__.py`).

So instead, I used old-fashioned unix to concatenate files:
```
cat SingleRegionAtlanticWTransportTransects.geojson Barents_Sea_Opening/transect.geojson Davis_Strait/transect.geojson Lancaster_Sound/transect.geojson Nares_Strait/transect.geojson  > SingleRegionAtlanticWTransportTransects_190730.geojson
```

<details>
<summary> Then I had to remove the extra little xml brackets in between: </summary>

**Replace**
```
        }
    ]
}
{
    "type": "FeatureCollection",
    "features": [
        {
```

**with**
```
        },
        {
```

</details>

Now I can use MpasMaskCreator to make the mask file:
```
MpasMaskCreator.x culled_mesh.nc  masks_SingleRegionAtlanticWTransportTransects.${MESHNAME}.${DATE}.nc -f SingleRegionAtlanticWTransportTransects_190730.geojson
```

And finally, a series of commands to visualize. I did a sum so all the transects are in one variable.
```
ncwa -y sum -v transectEdgeMasks -a nTransects masks_SingleRegionAtlanticWTransportTransects.oEC60to30J1.190715.nc masks_SingleRegionAtlanticWTransportTransects.oEC60to30J1.190715.sum.nc
cp initial_state.nc initial_state_sum_transects.nc
ncks -A -v transectEdgeMasks masks_SingleRegionAtlanticWTransportTransects.oEC60to30J1.190715.sum.nc initial_state_sum_transects.nc
vtk -f initial_state_sum_transects.nc -v transectEdgeMasks,temperature -d nVertLevels=0 --ignore_time -o vtk_oEC60to30J1.190715.sum
```
In Paraview, I visualize the masks as along edges and the temperature as a wireframe to show the primal cells:
image

mark-petersen:
Here is a zoom of the Lab Sea, and Greenland-Iceland-Norway:
image
It looks like the new transects and the process is working overall, but a few end points need to be put deeper into the land. The Greenland-Iceland transect angles strangely, and might need some way-points along the way.

milenaveneziani:
Thanks @mark-petersen: that's great.

A few comments from me:
1) It would be good to dig deeper and figure out how to use `merge_features` correctly. I know that we can add a `-f $featurefilename` option and it will then add that specific feature to the other standard transects. But I don't know how to do that with a bunch of new transects *at the same time*.
2) Sure, we can add a point inland on both sides of the transects to make sure that they are land-to-land. And we can update Fram Strait while we are at it.
3) How did you add the Atlantic transect, the one we use for the Atlantic MOC?

mark-petersen:
The Atlantic MOC transect was already there because I used unix cat to add it to the old file.  Here are a few more views:
image
image

milenaveneziani:
> The Atlantic MOC transect was already there because I used unix cat to add it to the old file

But I can't see the geojson feature for the SA transect in the geometric_feature repo. Do you have it somewhere else?

mark-petersen:
Huh! You are right. I'm glad you pointed that out. I've been using this file, but the MOC transect and region was not in geometric features:
https://gist.github.com/mark-petersen/8ae2aa8767e97abb6fb2afa82f0ea1da
image

milenaveneziani:
hmm, I see now. And this has both the transect and the region we use for the MOC calculation.

milenaveneziani:
ok, @mark-petersen : I have cleaned up some of the standard transects and improved a few of them (for example, for the Florida Strait, I made it go from Key West straight down to Cuba, without including any Gulf of Mexico section). I also added a lot of additional points for the long transects.
Please recompute the mask for the meshes that you'd like to test and let me know what you find.

As we discussed, we'll handle the South Atlantic transect business in a different PR.

geometric_data/ocean/transect/Africa-Ant/transect.geojson
```diff
@@ -17,14 +17,14 @@
                 "coordinates": [
                     [
                         30.000000, 
-                        -70.000000
+                       -70.000000
                     ], 
                     [
                         30.000000, 
-                        -31.300000
+                       -31.000000
```
milenaveneziani: I should add more points here.. Hold off just a minute

milenaveneziani:
ok, all done. please go ahead with testing @mark-petersen

milenaveneziani:
@mark-petersen: I found an example script that makes a full transect file. It's here, in the geometric_features repo:
`examples/setup_ocean_standard_transport_sections.py`
Problem is, I haven't yet figured out how to tell it *not* to download missing transects (because for example I don't want to include Robenson channel, which is in origin/master but *not* in my local branch). Will look at this more later.

milenaveneziani:
Pinging Andrew on this too: @proteanplanet

xylar:
@milenaveneziani and @mark-petersen, there is a lot to go through here.  I need to get myself caught up with the many issues brought up here, including making it clear how to use the local `geometric_data` rather than the repo.

I am concerned that at least one of the features being modified here is used in mesh generation so it is likely to be a problem to redefine it (Nares Straight).

A quick phone call next week (when I will be in AZ for a few days before heading over to NM) would probably help to clear some of this up.

mark-petersen:
@xylar thanks for noticing the Nares Straight. There are now two, one with a `Critical_Passage` tag and one with a `standard_transport_sections` tag.

mark-petersen:
I had to extend the Nares transport transect to get it to extend across. Here is the EC60to30 (Jigsaw version) with the transects:
image
Your additional points look good. Here is the current and my extended Nares:
image
If we go a bit inland, it doesn't really matter. I moved it north to avoid that channel in the high resolution meshes.

milenaveneziani:
> I am concerned that at least one of the features being modified here is used in mesh generation so it is likely to be a problem to redefine it (Nares Straight)

@xylar: can you please elaborate on that? How are the transect masks used in mesh generation?

@mark-petersen: I can easily remake the new Nares that goes more inland.

milenaveneziani:
OK, I have made a few more changes and then rebased to clean things up a bit. I added more points to Fram, 'cause I didn't like that zigzagging thing. And pushed a couple more transects more into land (Nares and LS).

geometric_data/ocean/transect/Nares_Strait/transect.geojson
```diff
@@ -1,42 +1,27 @@
...
-                "object": "transect", 
-                "component": "ocean", 
-                "author": "Xylar Asay-Davis", 
-                "depth": "250.0"
-            }, 
+                "name": "Nares Strait",
+                "tags": "standard_transport_sections",
+                "object": "transect",
+                "component": "ocean",
+                "author": "Milena Veneziani",
```
xylar: @mark-petersen, I believe you altered this feature so you should add yourself as an author.

```suggestion
                "author": "Milena Veneziani, Mark Petersen",
```

xylar: @milenaveneziani, are you okay with the movement of the transect a bit further north?

xylar:
>> I am concerned that at least one of the features being modified here is used in mesh generation so it is likely to be a problem to redefine it (Nares Straight)

> can you please elaborate on that? How are the transect masks used in mesh generation?

@mark-petersen already took care of this.  Several transects (with the `Critical_Passage` tag) are used to make sure there is ocean along the transect and that the ocean has at least some minimum depth.  There is also a "critical land blockage" transect for the Antarctic Peninsula that works as the inverse, ensuring that there is land everywhere along the transect.  The original Nares Strait feature was for this purpose but @mark-petersen has renamed this feature "Nares Strait deepen", so we should be fine.  The name isn't important, just the tag.

xylar:
@milenaveneziani, if you could run the following script:
``` python
import os
from geometric_features.utils import write_feature_names_and_tags


# Write a file features_and_tags.json with features and tags from the cache.
# This updates the file names, feature names and tags that geometric_features
# knows about.
write_feature_names_and_tags('./geometric_data')

# move features_and_tags.json into geometric_features to replace the old
# manifest
os.rename('features_and_tags.json',
          'geometric_features/features_and_tags.json')

# Make a geometric features object that gets data from a local cache in
# ./geometric_data.  (The remote branch won't matter.)
gf = GeometricFeatures(cacheLocation='./geometric_data')

# Make a feature colleciton with the standrd transport sections
fc = gf.read(componentName='ocean', objectType='transect',
             tags=['standard_transport_sections'])

# split the feature collection back into individual features within
# ./geometric_data to clean things up
gf.split(fc)
```
it will take care of my requested changes.  The script updates a [database](https://github.com/MPAS-Dev/geometric_features/blob/master/geometric_features/features_and_tags.json) that geometric_features uses to know what feature names and tags are valid.  It then merges all the features with the tag `'standard_transport_sections'` and splits them back out.  This will make sure formatting is kind of "standard".  It's not critical but it's meant to prevent features seeming like they're modified when there are no real changes.

xylar:
@mark-petersen wrote:
> I checked out this branch on grizzly and tried this:
> ```
> source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_compass.sh
> merge_features -c ocean -b transect  -t "standard_transport_sections" -o standard_transport_sections.geojson 
> ```

The problem is that `merge_features` is from the installation of `geometric_features` on conda-forge not the local git repo.  It has a "manifest" of all the features and tags appropriate to its version of geometric_features and they don't match the local `geometric_data`.  The way to do this locally is:
``` python

# Make a geometric features object that gets data from a local cache in
# ./geometric_data.  (The remote branch won't matter.)
gf = GeometricFeatures(cacheLocation='./geometric_data')

# Make a feature colleciton with the standrd transport sections
fc = gf.read(componentName='ocean', objectType='transect',
             tags=['standard_transport_sections'])

# Write the features
fc.write('standard_transport_sections.geojson')
```
but this will only work if the manifest gets updated in the local repo following the steps in my script from my previous comment.

It seems like the fact that I removed the local scripts `merge_features.py`, `split_features.py`, etc. is a problem so I can put stub scripts back in.

geometric_data/ocean/transect/Nares_Strait/transect.geojson
milenaveneziani: @xylar: I have lengthened the Nares transect inland by quite a bit, avoiding the small channel and without shifting it north. I have chosen the location of the transects to where I know there is observations to compare to. Before the final merge, it will be a good idea to re-create masks for a high res and a low mesh, just to make sure.
One other thing I wanted to say is whether we want to add some transects in the SO. I didn't do a thorough search for observations, but  the search I did do didn't produce many results.

milenaveneziani:
Thank you @xylar for providing the script to do this properly. The documentation I did read is the README file in the main repo directory. I neglected going into the `docs/`.
But I like how you have cleaned up everything and things are clear. Your new change to the documentation will also help in the future, I am sure (will look at the other PR asap).

geometric_data/ocean/transect/Nares_Strait/transect.geojson
xylar: Ah, got it.  So @mark-petersen's version is no longer there and he no longer needs to be a coauthor.  I think yours is the cleaner solution.

Regarding the SO, I'm not aware of good transects.  The best we're likely to get are a few moorings (e.g. at the Filchner Trough and various Amundsen Sea troughs) but I don't have a lot of time to figure out what the corresponding transects would be or to ask for the data right now.  Let's plan to add them later.

milenaveneziani:
@xylar, I ran the script you suggested, but I also had to edit a couple more things (it was getting confused that Nares Strait Deepen was still called Nares Strait, so I had to update that and the `features_and_tags.json` file).
I think now everything looks good. Let me know if there is anything else I missed.
I am going to recompute the masks for two meshes with these new transect features, just to make sure.

milenaveneziani:
OK, here are the standard transects using the EC60to30 mesh (left) and the ARRM60to10 mesh (right). As a reminder, there is no Nares Strait opening in the 60to30.

transects_labels

mark-petersen:
@milenaveneziani That looks great! Do you know what the blue vs red edges are showing?

I wonder if the northern endpoint for Florida-Cuba is appropriate. The original transect goes to Key West (from observational data), so presumably that captures the transport in the real ocean. But on a discrete grid, there may be eastward flow north of Key West.

Here is the current transect:
image

I think this would be safer. Two additional points are highlighted.
image

milenaveneziani:
> Do you know what the blue vs red edges are showing?

I plotted the EdgeSign, so red is positive and blue is negative.

Yes, I agree about the Florida-Cuba transect: I had noticed that too. Your suggestion looks great: I would only add a couple of extra points between Key West and -80.8,25.

mark-petersen:
Done. I just pushed four points, shown.
image

milenaveneziani:
Great, thanks @mark-petersen.
@xylar: when you have a minute (I know you are traveling right now), could you please give this a final check and approve if you are satisfied with the changes? Also, let me know if you'd like to discuss the MOC southern transect extractor (I left a comment on the issue that I closed last week, but I am not sure you saw that).

milenaveneziani:
Thanks @xylar!

xylar:
@milenaveneziani, could you run the merge/split script one last time to make sure nothing changes?  I think nothing will but just to be sure...

Then, feel free to merge.

We should discuss when to release a new version of geometric_features.  We should only wait if there are other features to add in the very near future.

milenaveneziani:
> We should only wait if there are other features to add in the very near future.

even if we get the MOC transect extractor to work, we don't need to add new features related to that, right?
And, we are not adding SO transects for now.
I believe @sbrus89 and/or Kristin (github name?) were thinking to add new transects for the North Atlantic: do @mark-petersen or @pwolfram have details on this?

xylar:
> even if we get the MOC transect extractor to work, we don't need to add new features related to that, right?

Yep, that's right.  The transects are gird-specific so don't make sense as geometric features.

> I believe @sbrus89 and/or Kristin (github name?) were thinking to add new transects for the North Atlantic: do @mark-petersen or @pwolfram have details on this?

Okay, I'll talk with @sbrus89, @kristin-hoch, @mark-petersen and @pwolfram before making a new release to make sure not to make extra work for myself.

PR #119, opened 2019-10-26
Add missing 'type' key to combined features

xylar:
Without this dictionary entry, the resulting feature is not recognized, e.g., by geojsion.io.

### DIFF
--- a/geometric_features/feature_collection.py
+++ b/geometric_features/feature_collection.py
@@ -220,6 +220,7 @@ class FeatureCollection(object):
                                  geometry['type']))
 
         feature = {}
+        feature['type'] = 'Feature'
         feature['properties'] = {}
         feature['properties']['name'] = featureName
         feature['properties']['component'] = \


xylar:
@milenaveneziani, I ran the following (similar to, but not identical to, what you had in your email):
```
./merge_features.py -c ocean -b region -n "Baffin Bay;Davis Strait" -o BaffinBaynew.geojson
./combine_features.py -f BaffinBaynew.geojson -n "Baffin Bay" -o BaffinBaynew2.geojson
```
Note: the merge command specifies the component and object type, which is required.  Using this branch, if I then view `BaffinBaynew2.geojson` at geojson.io, it looks seamless to me.  Could you try that out?

If you're still seeing gaps when you combine features, could you post an issue here with your steps so I can try to reproduce that problem.

milenaveneziani:
yes, this works for me too @xylar. Thanks!

Issue #118, opened 2019-10-24
Redefining Arctic and North Atlantic subpolar regions


milenaveneziani:
I see the following two main issues with the current definition of 'Arctic' and sub Arctic regions:
1) the current `Arctic` geojson feature is a weird region that covers most of the Arctic, but not quite all of it (excluding coastal seas such the Beaufort Sea, Chukchi Sea, etc);
2) the subpolar North Atlantic regions are defined weirdly sometimes, perhaps more in line with geographical definitions rather than oceanographic ones: one example is the Lab Sea, which is a small triangle, south of a `DavisStrait` feature..

Below, I am suggesting a possible redefinition. Please let me know if you have an opinion on this, and your suggestions, so that I can go ahead and make a PR to address it.

milenaveneziani:
**Arctic**

Currently, what we think of for 'Arctic' consists of the following geojson features:
1) Arctic
2) Barentsz (misspelled: should be simply `Barents`)
3) Kara Sea
4) White Sea
5) Laptev Sea
6) East Siberian Sea
7) Chukchi Sea
8) Beaufort Sea (*very* different from typical definition of Beaufort Gyre)
9) Lincoln Sea
10) and then there is a piece of the current Greenland Sea that should go in `Arctic`, north of Fram Strait (79N).

My first suggestion is to combine 1 and 5-10 above into a new `Arctic` feature.
My second suggestion is to combine 2-4 into a Barents-Kara Seas feature (this is because sometimes these two 'entrance' seas are considered separately from the main Arctic).
My third suggestion is to add new Arctic subregions, the details of which I still need to figure out, but they would approximately be the following:
- a Beaufort gyre (defined as from Proshutinsky's papers: 130W-170W, 70.5N-80.5N and deeper than 300m);
- a Eurasian Basin (deeper than ~1000m)
- a Amerasian Basin, excluding the Beaufort gyre piece (again deeper than ~1000m)
- a shelf region (shallower than ~1000m, expect for the BG piece again).

milenaveneziani:
# Subpolar North Atlantic

Here is what we have now:
1) Greenland Sea (mostly OK, but extends north of Fram Strait). Includes both Greenland and Iceland Seas
2) Norwegian Sea. Includes both Norwegian and Lofoten Basins
3) Baffin Bay
4) Davis Strait
5) Labrador Sea (different from conventional Lab Sea)
[...]

PR #123, opened 2019-10-30
Redefine Arctic ocean regions (Part 1)

milenaveneziani:
This is to completely revise the current Arctic ocean regions, which are mostly based upon geographic considerations, rather than oceanographic and sea-ice physical characteristics.

Intended to address #118.


milenaveneziani:
This PR is not ready, but I thought it would be a good idea to start opening it since the modifications involved are numerous..

To start things out, here are the new sub Arctic regions:
Screen Shot 2019-10-30 at 1 05 10 PM
These were added/modified **by hand**, although they were merged and split to get the associated geojson features in the right format.

A few things to note:
- Greenland Sea ends at Fram Strait, which is defined at 79N, consistently with our Fram Strait transect and observations available for Fram Strait;
-  Norwegian Sea ends at the Barents Sea opening, also consistent with our BSO transect;
- Labrador Sea and Baffin Bay meet at Davis Strait.

milenaveneziani:
For the Canadian Archipelago and Hudson Bay, I now have:

Screen Shot 2019-10-30 at 1 14 54 PM

after doing the following:
- combined Hudson Bay and Hudson Strait in one single feature;
- renamed the Northwestern Passages feature as 'Canadian Archipelago';
- at the moment, I left the Foxe Basin as part of the Canadian Archipelago feature, but please let me know if I should combine that with Hudson Bay.

milenaveneziani:
So now for the actual Arctic regions, I  am modifying the `arctic_ocean_regions.py` script to redefine them, but I am having trouble defining the Beaufort Gyre region, which has to be defined according to these constraints (from Proshutinski et al. 2009, doi:10.1029/2008JC005104):
- 70.5N-80.5N and 130W-170W, plus
- depths greater than 300 m

I am currently experiencing an error with the definition of the 300 m isobath contour (will point to the line where the error occurs in a minute).

Before getting into that, here is how I envision the 'Arctic Ocean', 'Barents Sea', and 'Kara Sea' features (note that I haven't changed them yet):

Screen Shot 2019-10-30 at 1 36 16 PM

feature_creation_scripts/arctic_ocean_regions/README.md
```diff
@@ -0,0 +1,23 @@
+# Arctic Ocean Regions
+
+TO BE EDITED
```
milenaveneziani: this file needs to be edited

feature_creation_scripts/arctic_ocean_regions/arctic_ocean_regions.py
```diff
@@ -0,0 +1,213 @@
...
+    epsilon = 1e-14
+    minY = numpy.amin(y)
+    wedge = shapely.geometry.Polygon([(epsilon, minY),
+                                      (epsilon**2, -epsilon),
+                                      (0, epsilon),
+                                      (-epsilon**2, -epsilon),
+                                      (-epsilon, minY),
+                                      (epsilon, minY)])
+
+    difference = poly.difference(wedge)
```
milenaveneziani: here is where I get an error, possibly because this code was originally written by @xylar to work with `bedmap` in the Southern Ocean. The error message is:
```
TopologyException: Input geom 0 is invalid: Self-intersection at or near point -1114116.2216271362 -2232860.7958771652 at -1114116.2216271362 -2232860.7958771652
  File "arctic_ocean_regions.py", line 224, in <module>
    main()
  File "arctic_ocean_regions.py", line 196, in main
    fcContour300 = get_longest_contour(contourValue=-300., author=author)
  File "arctic_ocean_regions.py", line 83, in get_longest_contour
    difference = poly.difference(wedge)
  File "/lcrc/group/acme/milena/miniconda3.7/envs/conda_geometric_features/lib/python3.7/site-packages/shapely/geometry/base.py", line 616, in difference
    return geom_factory(self.impl['difference'](self, other))
  File "/lcrc/group/acme/milena/miniconda3.7/envs/conda_geometric_features/lib/python3.7/site-packages/shapely/topology.py", line 70, in __call__
    self._check_topology(err, this, other)
  File "/lcrc/group/acme/milena/miniconda3.7/envs/conda_geometric_features/lib/python3.7/site-packages/shapely/topology.py", line 38, in _check_topology
    self.fn.__name__, repr(geom)))
shapely.errors.TopologicalError: The operation 'GEOSDifference_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at 0x2b28d45a0be0>
```

To try to debug, I plotted the result of the `cs` contour, and the two variables `poly` and `wedge`, which are all attached below, respectively:
cs_contour
poly
wedge

so it seems to me that there is no intersection between `wedge` and `poly`, but I  am not sure how to fix it..

milenaveneziani:
Pinging @proteanplanet (Andrew) here as I don't know how to add him as reviewer..

feature_creation_scripts/arctic_ocean_regions/arctic_ocean_regions.py
```diff
@@ -0,0 +1,213 @@
...
+    proshTags = 'Arctic;Proshutinsky'
+
+    # make a geometric features object that knows about the geometric data
+    # cache up a couple of directories
+    gf = GeometricFeatures('../../geometric_data')
+
+    fc = FeatureCollection()
+
+    # Combine old Barentsz_Sea and White_Sea into new Barents Sea 
+    # feature (Barentsz_Sea and White_Sea features will be removed)
```
milenaveneziani: this defines a new Barents Sea as a combination of Barentsz (?) and White Sea. I'd remove both of these features once this PR in..

feature_creation_scripts/arctic_ocean_regions/arctic_ocean_regions.py
```diff
@@ -0,0 +1,213 @@
...
+    fcBS.merge(gf.read('ocean', 'region', ['White Sea']))
+    fcBS = fcBS.combine('Barents Sea')
+    fcBS.to_geojson('BarentsSea.geojson')
+    fc = fcBS
+    props = fc.features[0]['properties']
+    props['tags'] = 'Barents_Sea;Arctic;Arctic_Basin'
+
+    # Include Kara Sea
+    fcKara = gf.read('ocean', 'region', ['Kara Sea'])
+    fc.merge(fcKara)
```
milenaveneziani: Kara Sea is left unchanged.

feature_creation_scripts/arctic_ocean_regions/arctic_ocean_regions.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Define full Arctic *but* Barents and Kara Seas
+    fcArctic.merge(gf.read('ocean', 'region', ['Arctic Ocean']))
+    fcArctic.merge(gf.read('ocean', 'region', ['Laptev Sea']))
+    fcArctic.merge(gf.read('ocean', 'region', ['East Siberian Sea']))
+    fcArctic.merge(gf.read('ocean', 'region', ['Chukchi Sea']))
+    fcArctic.merge(gf.read('ocean', 'region', ['Beaufort Sea']))
+    fcArctic.merge(gf.read('ocean', 'region', ['Lincoln Sea']))
+    fcArctic = fcArctic.combine('Arctic Ocean')
+    fcArctic.to_geojson('ArcticOcean.geojson')
```
milenaveneziani: write to geojson file is currently just a debugging line.

Note that I needed to add a triangle north of Fram Strait that is in the old Greenland Sea feature, but it really belongs to the Central Arctic.. (see `fc_tmp` lines above)

feature_creation_scripts/arctic_ocean_regions/arctic_ocean_regions.py
xylar: I see 2 problems.  The main one is that your contour hit s the edge of the boundary, unlike the one I was using.  You need a closed contour so you're going to have to fill in the bit that's getting connected with a straight line with two edges that follow the edge of you polar domain.  A self-intersecting shape is invalid according to the `shapely` library.

The much more minor problem is that your wedge is at 0 degrees longitude, not +/-180.  So it's going to divide your shape in a way we don't want.  Just make the y values positive instead of negative and you should be good.

xylar: One way to fix the contour might be to put topography values of zero all around the edges of your topography file on the polar grid.  That way, a -300 would also follow the edges of your plotting domain.

milenaveneziani: Thanks @xylar: all your suggestions worked.

milenaveneziani:
OK, I am done defining all the Arctic regions. Here is what I have, which reflects what I described at the top of this PR:

Screen Shot 2019-10-31 at 11 13 26 AM

Please, @proteanplanet and @maltrud (and maybe @vanroekel, if you have time), let me know what you think about the region definition (Andrew: you can quickly compare my map above with the MAISE regions; Mat: if you could take  a quick look at the maps of this PR and ping me if you  see anything you are uncomfortable with, I'd really appreciate it).

@xylar: any suggestion from you is of course welcome as well. The last bit I need to do is to update some of the features and perhaps preserve the old definition. First of all: is that something we want to do, rename say the 'original' Arctic_Ocean to 'Arctic_Ocean_Old', so that we can re-run the arctic feature script in the future if we want to? Or should we just get rid of the old Arctic_Ocean? Same goes with other features, such as Barentsz_Sea, Laptev_Sea, etc etc.

proteanplanet:
Hi @milenaveneziani, I'm struggling with the Beaufort Gyre and Central Arctic definitions. The Beaufort Sea sits right up against the Archipelago. If you are wanting to define the gyre itself, as you have named it, this does not define the right area, which can extend all the way into the Makarov Basin depending on the season, but nor is such a definition common in the literature because of the seasonality of the gyre. As a follow on, I suggest that the central Arctic definition should not include any part of the Canadian mainland or Banks Island coasts. More importantly, the central Arctic definition must strictly stop at Fram Strait, as per the attached figure and accompanying netcdf file, which was determined from ICESat masks commonly used for calculating ice fluxes.   Finally, I'd like to suggest that Archipelago, Baffin Bay, Greenland Sea or Labrador Sea definitions would be useful given the trouble in simulating these regions.

E3SM_icesate_mask

[ICESat_Masked_E3SM.nc.zip](https://github.com/MPAS-Dev/geometric_features/files/3795286/ICESat_Masked_E3SM.nc.zip)

milenaveneziani:
Hi Andrew, thanks for taking a look.

About the Beaufort Gyre, I am strictly following the definition of Proshutinski 2009 (see reference above): I'd like to keep that as is, for comparison with observations and previous model studies of BG freshwater content. If you like, we could add an additional definition of BG, perhaps later.

> I suggest that the central Arctic definition should not include any part of the Canadian mainland or Banks Island coasts

I am not seeing any essential difference between your attached plot and my proposed Canadian side of the Central Arctic. Perhaps I am missing something?

Also, keep in mind that we have a definition of `Arctic Ocean` that basically covers your blue area (I have indicated it in one of the maps of this PR).

> the central Arctic definition must strictly stop at Fram Strait
 
It does that now. Perhaps my plot is confusing?

> I'd like to suggest that Archipelago, Baffin Bay, Greenland Sea or Labrador Sea definitions would be useful

I have included pictures of all these regions in the first part of this PR. Please take a look at that and let me know what you think (I did ask a question about Hudson Bay and the Archipelago at one point).

milenaveneziani:
Here is our `Arctic Ocean` as is now (sum of all the regions plotted in the previous map, *minus* Barents and Kara Seas):

Screen Shot 2019-10-31 at 12 48 40 PM

xylar:
> First of all: is that something we want to do, rename say the 'original' Arctic_Ocean to 'Arctic_Ocean_Old', so that we can re-run the arctic feature script in the future if we want to? Or should we just get rid of the old Arctic_Ocean? Same goes with other features, such as Barentsz_Sea, Laptev_Sea, etc etc.

Oh, no, we definitely just want to get rid of the old ones.  Keeping them around would be confusing.  People can check out old tags of this repo if they want the old geometry.

proteanplanet:
@milenaveneziani, if you want to adhere to Proshutinski 2009, I have no objection, but let's be clear that that definition is not commonly used, nor does it represent the Beaufort Gyre.  In fact, it seldom includes the center of the gyre. Regardless, all divisions of the Arctic Ocean are a little contrived, including MASIE, so it's up to you if this helps your research.  I still urge strict adherance to the Fram Strait, rather than the latitudinal approximation you have now, otherwise it simply can't be used to sufficiently compare with most observations of Arctic outflow and sea ice mass from the existing literature. I personally will have to run my own diagnostics and won't be able to use it in its current form.

milenaveneziani:
@proteanplanet, I think we can have two definitions of BG, no question about that. If you have a suggestion for lon,lat values for a second BG, I can add it now, but if you don't have time at the moment we can also add it later (I can open an issue so that we remember).
About Fram Strait: I can certainly look at the netcdf file you attached (although if you have a list of lon,lat values, it would be faster for me). I had chosen 79N previously for the FS transect because I have found a couple of papers on observations at 79N across the Strait. But to me an approximation of 79-80N is good enough, so I can also use the one in the ICESat file.

Both you and Wieslaw have raised questions about the Canadian side of the Central Arctic, so I will communicate with both of you about that by email.

proteanplanet:
Would a shapefile be best?

milenaveneziani:
A quick google search told me that I can convert a shapefile to geojson, so yes, I think that would be helpful. thanks.

milenaveneziani:
After conversations  with both Andrew and Wieslaw (Maslowski), I'd like to split this work in two parts, if you don't mind @xylar. This is because we need to define two sets of regions: one that is better suited for sea-ice and follows the NSIDC regions, and one set that is better suited for ocean (for which, for example, the shelves are separated from the deep part of the regions). I would like to assign two separate tags to these sets of regions: `Arctic_NSIDC` and `Arctic`.
Right now this PR has a mix of them.

The reason why I'd like to merge this as PART I is that 1) it got a little too big, and 2) I have some other things that I need to figure out and that it may be better to address in a separate PR. One of them is, for example, why I cannot access the local cache for say the Canadian  Archipelago (renamed from Northwestern PAssages). For some reason, I am getting an error on that.

Anyway, please @xylar take a look at this and let me know if there is anything I should adjust for the current regions. I'll  be checking this again later today. Thanks.

xylar:
@milenaveneziani, I'm fine with that division.  I don't have strong opinions about the regions.

If you want to post some more info or help me reproduce, I'm happy to help you debug the local cache issue.  Off the top of my head, the most likely issue is that the database of tags and features hasn't been updated to include this feature in the branch you're testing.  Or it could somehow be trying to download features from 1.0.4 tag if they don't exist as it expects in your local cache (e.g. file name isn't quite what it's expecting).

Happy to try to help.

xylar:
@milenaveneziani, I ran the following script and I get an error that one of the features is self-intersecting.  I'm going to try to figure out which.
```
#!/usr/bin/env python

from geometric_features import GeometricFeatures, FeatureCollection

gf = GeometricFeatures('./geometric_data')

fc = FeatureCollection()
fc.set_group_name(groupName='OceanBasinRegionsGroup')

# build ocean basins from regions with the appropriate tags
oceanName ='Arctic'
basinName = '{}_Basin'.format(oceanName)
print(oceanName)

print(' * merging features')
fcBasin = gf.read(componentName='ocean', objectType='region',
                  tags=[basinName])

print(' * combining features')
fcBasin = fcBasin.combine(featureName=basinName)

fcBasin.to_geojson('arctic_combined.geojson')
```
Result:
```
Arctic
 * merging features
 * combining features
TopologyException: Input geom 1 is invalid: Self-intersection at or near point 18.214821807130939 80.100459032501917 at 18.214821807130939 80.100459032501917
  File "./combine_arctic.py", line 20, in <module>
    fcBasin = fcBasin.combine(featureName=basinName)
  File "/home/xylar/code/mpas-work/geometric_features/redefine_arcticRegions/geometric_features/feature_collection.py", line 209, in combine
    combinedShape = shapely.ops.cascaded_union(featureShapes)
  File "/home/xylar/miniconda3/envs/geometric_features/lib/python3.7/site-packages/shapely/ops.py", line 131, in cascaded_union
    return geom_factory(lgeos.methods['cascaded_union'](collection))
  File "/home/xylar/miniconda3/envs/geometric_features/lib/python3.7/site-packages/shapely/geometry/base.py", line 76, in geom_factory
    raise ValueError("No Shapely geometry can be created from null value")
ValueError: No Shapely geometry can be created from null value
```

This will likely cause us trouble when we're trying to create basin masks so needs to be fixed.

xylar:
## Update

I ran:
```python
#!/usr/bin/env python

import shapely.geometry

from geometric_features import GeometricFeatures, FeatureCollection

gf = GeometricFeatures('./geometric_data')

fc = FeatureCollection()
fc.set_group_name(groupName='OceanBasinRegionsGroup')

# build ocean basins from regions with the appropriate tags
oceanName ='Arctic'
basinName = '{}_Basin'.format(oceanName)
print(oceanName)

print(' * merging features')
fcBasin = gf.read(componentName='ocean', objectType='region',
                  tags=[basinName])

for feature in fcBasin.features:
    featureName = feature['properties']['name']
    shape = shapely.geometry.shape(feature['geometry'])
    print('{} is_valid: {}'.format(featureName, shape.is_valid))

print(' * combining features')
fcBasin = fcBasin.combine(featureName=basinName)

fcBasin.to_geojson('arctic_combined.geojson')
```
and I see:
```
$ ./combine_arctic.py 
Arctic
 * merging features
Arctic Ocean is_valid: False
Barents Sea is_valid: True
Beaufort Gyre is_valid: True
Beaufort Sea is_valid: True
Canadian Archipelago is_valid: True
Central Arctic is_valid: False
Chukchi Sea is_valid: True
East Siberian Sea is_valid: True
Kara Sea is_valid: True
Laptev Sea is_valid: True
Lincoln Sea is_valid: True
White Sea is_valid: True
...
```
so it looks like the Central Arctic is the problem feature.  I'll look further.

xylar:
This weird line or features similar to it might be the problems:
image

I recall a trick where I use the outline of the shape from shapely to remake the shape, but I'll need to investigate.

xylar:
Here's a code for fixing the feature:
``` python
#!/usr/bin/env python

import shapely.geometry

from geometric_features import GeometricFeatures, FeatureCollection

gf = GeometricFeatures('./geometric_data')

fc = gf.read(componentName='ocean', objectType='region',
             featureNames=['Central Arctic'])
for feature in fc.features:
    featureName = feature['properties']['name']
    shape = shapely.geometry.shape(feature['geometry'])
    print('{} is_valid: {}'.format(featureName, shape.is_valid))
    if not shape.is_valid:
        fixed = shape.buffer(0)
        print('  Fixed? {}'.format(fixed.is_valid))
        feature['geometry'] = shapely.geometry.mapping(fixed)
fc.to_geojson('central_arctic_fixed.geojson')
```
In general, this `buffer(0)` trick is pretty handy: https://shapely.readthedocs.io/en/stable/manual.html#object.buffer

And a plot of the result:
image

milenaveneziani:
Thanks @xylar for checking the regions and suggesting  the fix to the Arctic/Central Arctic. I tried to implement the fix, but I run into the same problem again. I try to read in my newly defined Central_Arctic region from my local cache location, and I get this error:
```
  File "arctic_ocean_regions.py", line 314, in <module>
    main()
  File "arctic_ocean_regions.py", line 287, in main
    fcCentralArctic = gf.read('ocean', 'region', ['Central Arctic'])
  File "/global/project/projectdirs/acme/software/anaconda_envs/cori/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/geometric_features/geometric_features.py", line 103, in read
    featureNames, tags, allTags)
  File "/global/project/projectdirs/acme/software/anaconda_envs/cori/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/geometric_features/geometric_features.py", line 271, in _get_feature_names
    featureName))
KeyError: 'invalid feature Central Arctic'
```
I really don't understand why it complains since 1) I do point to my local cache and 2) I checked `geometric_features/features_and_tags.json` and that is updated (you can see it in this PR too), with what I think are the right names of the features.

milenaveneziani:
I tried this with both my own  conda and also with your latest e3sm-unified (which is working fine for me on  cori, btw).

xylar:
@milenaveneziani, I think I see what's happening.  You are using the package from your conda environment (including its cache of tags and feature names) rather than the local version.  I will see if I can figure out an easy solution.

xylar:
@milenaveneziani, I'm able to reproduce the problem if I put my script `fix_central_arctic.py` (the last one above) in a directory other than where the `geometric_features` and `geometric_data` directories from this branch are located.  But if I run with the script in the root directory of this branch, it picks up the local version of the python package rather than the conda package from the environment and everything runs fine.

One thing you could do for "safer" testing would be to make a conda environment with all of the requirements for `geometric_features` without the package itself:
```
conda create -n geometric_features -c conda-forge python=3.7 numpy matplotlib \
    cartopy shapely requests progressbar2
```

milenaveneziani:
Hi @xylar: I made the geometric_features conda env, but I still get an error when running the script in the subdirectory. It does work for me too if I run it in the main geometric_features/ location.

milenaveneziani:
Anyways, the new Arctic_Ocean and Central_Arctic features are fixed, although I still noticed that detached piece that should really be just part of the Kara Sea:
Screen Shot 2019-11-02 at 10 36 50 AM
Also, if you look closely at the Arctic_Ocean feature, there are places in the middle of it with tiny islands (probably a result of combining different features).
Wouldn't it be better to have a smoother polygon when creating our masks?

xylar:
> Hi @xylar: I made the geometric_features conda env, but I still get an error when running the script in the subdirectory. It does work for me too if I run it in the main geometric_features/ location.

This is to be expected.  Python has to find the `geometric_features` package relative to the directory where you're running the script.  This means your script can't be in a subdirectory unless you link the `geometric_features` path into that subdirectory.

In examples, I have these instructions:
"To use these scripts, you should either install the geometric_features conda package (see README.md in the root of the repo) or copy the scripts into the root directory."
If you have altered features, you can't use the conda package because it doesn't include the new features, so only the option of running from the root directory remains.

xylar:
> Anyways, the new Arctic_Ocean and Central_Arctic features are fixed, although I still noticed that detached piece that should really be just part of the Kara Sea

I'm not worried about that bit because it's basically just land either way.

> Also, if you look closely at the Arctic_Ocean feature, there are places in the middle of it with tiny islands (probably a result of combining different features).
Wouldn't it be better to have a smoother polygon when creating our masks?

You mean without the coastline getting cut out?  Yeah, probably that would ideally be the case.  That way, you don't have the risk that cells happen to fall between the cracks of the ocean basin and land masks when they don't quite match up.  But it's also probably too much trouble to try to expand all the ocean features that are already defined.

milenaveneziani:
> This is to be expected. Python has to find the geometric_features package relative to the directory where you're running the script. This means your script can't be in a subdirectory unless you link the geometric_features path into that subdirectory.

So, you did not use the antarctic script in the subdirectory that it is now? But then why specify the local cache directory in the `gf` feature collection?

OK, I am glad you are not worried about the extra piece of land.

What about these tiny islands in the Siberian side of the Arctic?:
Screen Shot 2019-11-02 at 11 12 18 AM

I agree that it would be a huge pain to redefine everything..

xylar:
> So, you did not use the antarctic script in the subdirectory that it is now? But then why specify the local cache directory in the gf feature collection?

The purpose of specifying the cache directory is so you can make `geometric_features` download whatever it needs to that cache directory.  The location of `geometric_data` is unrelated to the location of `geometric_features`, and the latter is what knows about what features are available (via the database).

I sometimes run scripts in a subdirectory (similarly in `mpas_analysis`) but I always have to create a local link to the python package when I do that.  `mpas_tools` is full of such symlinks, which is not great but seems to be the only way to reconcile the old script-based way of doing things with the newer package method.

xylar:
> OK, I am glad you are not worried about the extra piece of land.
> 
> What about these tiny islands in the Siberian side of the Arctic?

One option would be to include any islands in any seas (but obviously not the ones on the boundary between seas).  A trick in `shapely` for doing this is:
``` python
no_holes = MultiPolygon(Polygon(p.exterior) for p in m)
```
https://stackoverflow.com/a/48122325

milenaveneziani:
Regions should be all fixed now. And thanks for explaining the cache location and process. I have now included a local link to the python main directory for geometric_features.

I am satisfied with this at the moment, but I will add a Part II, with some of the regions better defined for the ocean, sometime next week.

xylar:
We can make a new release whenever you want but it might be simpler to wait until part 2 of your changes are done if you think that will happen soon.

milenaveneziani:
Great, thanks so much for the weekend review!

Could you please elaborate on how I do this?:
> To use the new version once this is merged, you'll have to specifically point your `GeometricFeatures` object to `master`

And yes, I will do part II as soon as I can.

xylar:
> Could you please elaborate on how I do this?:
>> To use the new version once this is merged, you'll have to specifically point your GeometricFeatures  object to master

Sure!  You won't be able to use the scripts like `merge_features.py` as they are because they pull features from the tag of the release given in `__init__.py` and there's not (yet) a command-line option to alter this.

In a python script, you'd need to do something like:
``` python

# get geometric data from geometric_features master branch and store it in
# the local directory ./geometric_data
gf = GeometricFeatures(localCache='./geometric_data',
                       remoteBranchOrTag='master')

# read in a feature collection defining the coastline
fc = gf.read(componentName='natural_earth', objectType='region',
             featureNames=['Land Coverage'])
```

Here's the [relevant documentation](https://geometric-features.readthedocs.io/en/latest/geometric_features.html).

milenaveneziani:
Thanks @xylar!

PR #125, opened 2019-11-05
Redefine Arctic regions part II

milenaveneziani:
This completes the definition of Arctic regions started in part I (#123). 

We now have 3 different tags for Arctic regions:
1. `Arctic`: oceanography-relevant regions (where, for example, shelves are separated from deep regions);
2. `Arctic_NSIDC`: seaice-relevant regions, following the definition found here:
https://nsidc.org/data/masie/browse_regions
https://nsidc.org/data/g02186?qt-data_set_tabs=2
3. `Arctic_Proshutinsky`: this contains the Beaufort Gyre as defined in https://doi.org/10.1029/2008JC005104, and associated Beaufort Gyre Shelf.


milenaveneziani:
Here is a plot of the `Arctic`-tag regions:

Screen Shot 2019-11-04 at 5 04 54 PM

milenaveneziani:
Here is one of the `Arctic_NSIDC` regions:

Screen Shot 2019-11-04 at 4 27 14 PM

milenaveneziani:
And finally, here are the two `Arctic_Proshutinsky` regions:

Screen Shot 2019-11-04 at 4 44 04 PM

milenaveneziani:
@proteanplanet: the `Arctic_NSIDC` regions should satisfy your request about the Beaufort Sea, Central Arctic, and particularly, Fram Strait.

proteanplanet:
@milenaveneziani Thanks for doing this, it is great.

geometric_features/features_and_tags.json
```diff
@@ -932,6 +932,11 @@
         "Arctic",
         "Atlantic_Basin"
       ],
+      "Baffin Bay NSIDC": [
+        "Baffin_Bay_NSIDC",
+        "Arctic_NSIDC",
+        "Arctic_Basin"
```
xylar: @milenaveneziani, I believe the main way we use the "Arctic_Basin" tag is to pull out the features that need to be `combine`d to produce the single feature that is the `Arctic_Basin`.  Would it make sense to only tag the "oceanographic" features for this purpose?  This is how the other ocean basins work.

milenaveneziani:
While checking Arctic_Basin and Atlantic_Basin tags, I realized I had upset some of the Atlantic regions in such a way that combining the Atlantic_Basin features would produce a self-overlapping area. So I adjusted the Lab Sea and North Atlantic Ocean features in such a way that this no longer happens.

Now, the Arctic_Basin tag should produce a decent Arctic Ocean, and the Atlantic_Basin tag should produce a decent Atlantic Ocean (the latter has some seams that should be fixed before we create a mask, and which are independent from the changes included in this PR).

milenaveneziani:
Thanks @xylar for the review and the help on these!

PR #127, opened 2020-03-03
Suppress progress bar in differencing features

xylar:
Progress is not shown by default.

This progress is showing up in MPAS-Analysis runs that create the MOC regions, and it isn't convenient in that context because it does not go to a log file and instead interrupts the task progress bar.


xylar:
## Testing

Ran `setup_ocean_region_groups.py` from examples and no longer see progress during masking of the MOC basins.

PR #128, opened 2020-03-03
Update to version 0.1.6

xylar:
Get version number from `geometric_features/__init__.py` instead of having to change it everywhere (following the approach used by MPAS-Tools).

Remove conda recipe, since we want to use the conda-forge feedstock instead.


xylar:
A test build with the conda-forge recipe was successful.

PR #129, opened 2020-03-04
Add GEOMETRIC_DATA_DIR environment variable

xylar:
This is the default location to look for geometric data if it is set.  The ability to set this variable will enable optionally packaging the data with the conda package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,29 @@ found in the `examples` directory.  Each of the classes and functions that make
 up the package have extensive documentation.  More user-level documentation
 will be added shortly.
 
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I made a conda package with this environment variable pointing to the offline data (see https://github.com/conda-forge/geometric_features-feedstock/pull/9).  I used this package to run `setup_ocean_region_groups.py` from examples but without pointing to `./geometric_data`.  I ran this script in its own directory to make sure it wasn't picking up local geometric data.  Everything ran as expected.

PR #130, opened 2020-03-11
Add Travis CI for testing package and building docs

xylar:
The approach is similar to MPAS-Tools in that it builds the conda package similarly to conda-forge, and then used the package to run test and build the documentation.


xylar:
@pwolfram, this is more or less the same as https://github.com/MPAS-Dev/MPAS-Tools/pull/301 but for `geometric_features`.  I just want to make sure you don't spot any major problems.  I'll merge tomorrow but do let me know what you think if you have time to look at this today.

PR #136, opened 2020-04-16
Add Black Sea blockage

xylar:
This feature makes sure that the Black Sea is not included in MPAS meshes (as long as they use critical passages and critical land blockages, which high resolution meshes typically do not).

### DIFF
--- a/geometric_data/ocean/transect/Black_Sea_Blockage/transect.geojson
+++ b/geometric_data/ocean/transect/Black_Sea_Blockage/transect.geojson
@@ -1,0 +1,37 @@
+{
+    "type": "FeatureCollection",
+    "features": [
+        {
+            "type": "Feature",
+            "properties": {
+                "name": "Black Sea Blockage",
+                "tags": "Black_Sea_Blockage;Critical_Land_Blockage",
+                "object": "transect",
+                "component": "ocean",
+                "author": "Xylar",
+                "height": "100.0"
+            },
+            "geometry": {
+                "type": "LineString",
+                "coordinates": [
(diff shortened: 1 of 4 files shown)


xylar:
@maltrud, @milenaveneziani, @jonbob and @mark-petersen, just take a look at this:
https://github.com/xylar/geometric_features/blob/add_black_sea_blockage/geometric_data/ocean/transect/Black_Sea_Blockage/transect.geojson
and let me know if you agree that this should be blocked off.  Nothing else needed for your review for now.

jonbob:
@xylar - does it also get rid of the Caspian?

xylar:
I don't see any way the Caspian Sea would ever connect to the open ocean. If it were through the Black Sea, then yes this would block it, too. Do you know of meshes where it wasn't culled?

jonbob:
I think it's always been through the Black Sea, so this should take care of it. I just wasn't picturing it from the blockage transect

xylar:
@milenaveneziani and @maltrud, do you want to review this?  It takes quite a bit of work to update geometric_features and compass so these changes make it into meshes, and I'd like to get started on that as soon as I can.

xylar:
@vanroekel, would you also give it a look?

xylar:
I did a run with the meshes from https://github.com/MPAS-Dev/MPAS-Model/pull/522 and https://github.com/MPAS-Dev/MPAS-Model/pull/518 using a compass environment with this version of geometric features.  I didn't verify yet that everything worked as expected but neither mesh has the Baltic Sea...

PR #140, opened 2020-06-18
Rename Ross East --> Western Ross...

### LINKED ISSUE #139: Ross Ice Shelf melt time series labels
Regional labels for the Ross Ice Shelf sectors should be changed from the Rignot et al. labels "Ross West" and "Ross East" to "eastern Ross" and "western Ross," respectively, or similar.

xylar:
...and similarly Ross West --> Eastern Ross.  Rignot et al. (2013) named these parts of the shelf based on which hemisphere or ice sheet they're in, but this is very confusing given that the Ross
crosses the antimeridian and thus its eastern side is in West Antarctica and its western side in East Antarctica.  To be true to the local cardinal directions, we're renaming the halves in the more standard way.

closes #139


xylar:
@stephenprice, @matthewhoffman, @darincomeau and @cbegeman, you can just review by inspection.  Let me know if you have questions but the main thing is to make sure we don't get so confused about which half is the eastern vs. the western half.

For this to propagate to MPAS-Analysis, I'm going to have to make up a new name for the collection of regions (e.g `iceShelvesRev2`) and generate new cached mask files for our standard meshes.  I'll do that once this gets merged and we have a new version of `geometric_features` to work with.

xylar:
A plot of each part and the combined Ross, just to make sure nothing is messed up:

Figure_1
Figure_2
Figure_3

cbegeman:
Looks great, @xylar. Thanks!

stephenprice:
This is fine with me. In fact, it's always driven me a little crazy that it was named like this, having finally (only after 20 yrs) trained my brain to determine east / west directions when looking at maps of Antarctica. Thanks for fixing it.

xylar:
@matthewhoffman, I'm taking you off as a reviewer because I think that's enough eyes on this.  But not because I don't value your opinion.  Feel free to weigh in after the merge.

matthewhoffman:
@xylar , thanks - looks good.

PR #142, opened 2020-07-07
Remove problematic characters from feature dirs

xylar:
This merge removes characters that are sometimes problematic from the directory names of features.

This merge also fixes some incorrect docstrings, adding a missing component and correcting the use of tags for reading features.

### DIFF
--- a/geometric_features/geometric_features.py
+++ b/geometric_features/geometric_features.py
@@ -80,8 +80,8 @@ class GeometricFeatures(object):
 
         Parameters
         ----------
-        componentName : {'bedmap2', 'iceshelves', 'landice', 'natural_earth', 'ocean'}
-            The component from which to retieve the geometric features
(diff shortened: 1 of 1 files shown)


xylar:
@proteanplanet, this should make it safe to use the feature names you originally had, in case you want to do that.

xylar:
## Testing

I merged all the features and split them again, proving that none of the existing names is affected by removing these potential problem characters.  Here was my script:

```python
#!/usr/bin/env python
from geometric_features import GeometricFeatures, FeatureCollection


gf = GeometricFeatures(cacheLocation='./geometric_data')

fc = FeatureCollection()
for component in ['bedmachine', 'bedmap2', 'iceshelves', 'landice',
                  'natural_earth', 'ocean']:
    for objectType in ['point', 'transect', 'region']:
        if objectType in gf.allFeaturesAndTags[component]:
            print(component, objectType)
            fc.merge(gf.read(componentName=component, objectType=objectType))


gf = GeometricFeatures(cacheLocation='./new_geometric_data')
gf.split(fc)
```

PR #143, opened 2020-07-08
Addition of Arctic shipping routes

proteanplanet:
This adds in five shipping routes - four Northwest Passage variants and one Northern Sea Route - to the Arctic, as shown in the figure below.

E3SM_Forced_Arctic_Passages


xylar:
@vanroekel, @mark-petersen and @maltrud, could you take a look at these and let me know if they look good to you?

mark-petersen:
@xylar, these files are large, 500-600k, with 4000 to 5000 points each. Thanks for testing. It will be good to see if that causes any problems, and if our python scripts can handle this in a reasonable amount of time. We have not run critical passages with this size files before.

@proteanplanet what tool did you make these in? It must be an automated tool - what was your choice for the step in distance?

One way to at least make the files smaller is to avoid the depth line in each point:
```
                    [
                        -169.023116,
                        59.291589,
                        0.0
                    ],
```
so that the line with 0.0 does not show. Was that an option in your tool?

xylar:
@mark-petersen, I don't think the file size is an issue but I agree that the number of points is much higher than is likely needed even a high resolution.  I hadn't noticed the 3rd coordinate.  We should remove those if possible because no other features have them.

mark-petersen:
The number of points is only a problem if it causes a significant slow-down in the mesh generation, so we'll see if it is an issue.

xylar:
Hmm, my timing suggests that this is going to have a bad impact on our nightly regression testing.
Timing before these transects were added
```
$ time ./run.py 
real	0m52.563s
user	0m45.359s
sys	0m3.351s
```

Timing with these transects included:
```
$ time ./run.py
real	4m3.028s
user	3m23.593s
sys	0m3.515s
```

So the entire `cull_cells` step is taking more than 4x as long.  I wouldn't worry about this for any other test cases but for QU240, we'd like it to be fast so regression testing doesn't slow down the whole development process.  One option would be to turn off critical passages in the QU240 setup or to explicitly exclude these transects from that configuration.

@proteanplanet, presumably an alterative would be to reduce the number of points considerably (say by a factor of 10).  How hard would that be and how much might that impact the fidelity at higher res?

proteanplanet:
> @proteanplanet what tool did you make these in? It must be an automated tool - what was your choice for the step in distance?
> 
> One way to at least make the files smaller is to avoid the depth line in each point:
> 
> ```
>                     [
>                         -169.023116,
>                         59.291589,
>                         0.0
>                     ],
> ```
> 
> so that the line with 0.0 does not show. Was that an option in your tool?

@mark-petersen This is the development chain: 1) I generated the ship routes using Arctic Council transects as the initial template. My coding included evenly discretizing the tracks at 1 nautical mile (great circle) spacings.  2) From this I generated a KML file and then ingested that into https://geojson.io.  The KML step is (unnecessarily) adding the altitude.

@xylar Now we know that this will affect performance, I can very easily fix this development chain. 1) Reduce the transects to only the narrow passages, probably resulting in 90% saving and removing the altitude. 2) Directly code the GeoJSON to avoid step 2 entirely above. 

I'll regenerate.  I can close this PR, and reissue another one, with short transects.

xylar:
Okay, sounds good.

PR #145, opened 2020-07-08
include transects important for Arctic shipping

proteanplanet:
This adds nine critical transects from five shipping routes - four Northwest Passage variants and one Northern Sea Route - to the Arctic, as shown in the figure below.

E3SM_Forced_Arctic_Passages_2


xylar:
## Testing

I timed the `culled_mesh` step of the QU240 init test case:
```
real	1m51.921s
user	0m45.522s
sys	0m2.477s
```
This is comparable to previously:
```
$ time ./run.py 
real	0m52.563s
user	0m45.359s
sys	0m3.351s
```
 and seems fine to me.

I'll run on higher resolution test cases to see what the results look like.

proteanplanet:
@xylar  Thanks for running the tests.  Is it possible for us to only apply these passages for ~15km resolution and higher?  So they would apply for WC14, but not EC30_60?

xylar:
@proteanplanet, no, that's not currently possible but it wouldn't be too difficult to change.  We would need a different tag instead of `Critical_Passage` and we would need a different flag within COMPASS for signalling these transects are also needed.  This wouldn't be difficult.

Is your concern about computational speed, I think we're fine.  If your concern is that these channels might be too much on the `EC30to60km` grid, I think it would be better to at least see what the effect is before we make extra work for ourselves.

proteanplanet:
> Is your concern about computational speed, I think we're fine. If your concern is that these channels might be too much on the `EC30to60km` grid, I think it would be better to at least see what the effect is before we make extra work for ourselves.

Yes, let's see what it looks like first.  

It matters most how fluxes will change, and we aren't deepening the narrows, so these might now change that much.   The coastline may not look realistic at standard resolution, but that might not matter.  e.g. Prince of Wales Strait should strictly only be open at ~6km as it was in E3SM-HR V1, however, I dug up a paper suggesting it's likely to become the most prevalent shipping channel on the Canadian side, now becoming ice-free in summer more rapidly than any of the other passages we're opening up.  But let's see where it lands us.

xylar:
## Results

### QU240
QU240

### EC30to60
EC30to60km

### WC14
(Coming soon!)

### ARM10to60
(Coming soon!)

xylar:
Oops, I made a mistake in the timing above.  That was without critical passages.  I need to update.

proteanplanet:
Wow. EC30to60 looks really nice.

proteanplanet:
Maybe assign Milena as a reviewer, too?

xylar:
@mark-petersen, @maltrud and @vanroekel, could you take a look at these an approve if you're happy?  This can just be by inspection and my plots above if you like.

Once it's approved, I'll make the necessary changes to do a `geometric_features` release and a new `compass` metapackage, and then to update COMPASS on MPAS-Model.

xylar:
Sorry, of course @milenaveneziani should be a reviewer!

proteanplanet:
_Background information for documentation purposes only_: The transects are respectful of most small islands important for Arctic fluxes, which is not immediately apparent in the heading figure. For example, the Bering Strait transect routes around the Diomede Islands, important for the Pacific inflow. However, the transects are deliberately less respectful of a few similarly-sized small islands in the Archipelago, so that we don't artificially lock up these channels with grid-enforced land-fast ice at 5km resolution and above.  Therefore, generating a new E3SM-HR 6-18km mesh with Jigsaw, if it were to happen, would provide further enhancements over the existing high-resolution mesh.  Further exploration can be done in Google Earth with this file, that includes the transects added in this PR: 

[E3SM_Ship_Channels.kml.zip](https://github.com/MPAS-Dev/geometric_features/files/4898951/E3SM_Ship_Channels.kml.zip)

milenaveneziani:
This looks great to me. I did see it this morning, but had not commented 'cause I needed to leave for Albuquerque.

So, for what I understand, the channel opening is done right after mesh creation, and it therefore only affects land-sea masking (or cell culling, in MPAS lingo?). Then, after this, comes the add-topography step, and I suppose that the channel bathymetry is then regulated by resolution of the mesh as well as the resolution of the topography data. Does that sound about right?

milenaveneziani:
@proteanplanet: so, did you use Google Earth data for defining the transects?
I would like to add one for the Foxe Basin, which I don't think we have in the current critical_passages list.

proteanplanet:
> @proteanplanet: so, did you use Google Earth data for defining the transects?
> I would like to add one for the Foxe Basin, which I don't think we have in the current critical_passages list.

Google Earth was not used to generate these paths; Keyhole Markup was only provided here for quick exploration. The base shipping paths derive from a digital appendix of the [Arctic Marine Shipping Assessment (2009)](https://pame.is/images/03_Projects/AMSA/AMSA_2009_report/AMSA_2009_Report_2nd_print.pdf) and have been further edited and updated in consultation with PNNL shipping group. I coded the AMSA paths and changes based on previous experience in ship routing. All course changes follow great circles at an along-track resolution of 1 nautical mile. These paths are part of a larger set of tracks being tested across the Arctic for economic and met-ocean viability, including those shown here:

InteRFACE_Shiptracks

I have no objection with you adding the Foxe Basin for WC14 r03, but that is not a ship route, and should really be in a different PR.  I can work that up right now, if you like.

milenaveneziani:
Thanks @proteanplanet. No, I didn't mean to include the Foxe Basin in this PR, it really does not belong here. I can do it in a different PR when I get some time.

mark-petersen:
@xylar thanks for testing. Those plots look great!

@proteanplanet you wrote "I coded the AMSA paths". Does that mean that you wrote some matlab code to produce the geojson files? Regardless of the method, could you see if you can easily change your process to remove the 0 z-coordinate from every point?

xylar:
@mark-petersen, the z-points are already gone.  Where are you seeing them?

xylar:
@mark-petersen, this is an entirely new PR from the one with z values. This has new, shorter transects. The one you were referring was #143 and has been closed. I hope that clears things up.

mark-petersen:
I see now. Thanks! And @proteanplanet thanks for the revision.

PR #146, opened 2020-07-10
Add Foxe Basin Throughflow

proteanplanet:
This pull request is at the suggestion of @milenaveneziani to include the Foxe Basin Through Flow in addition to the Arctic ship routes for WC14 r03, and is upstream of the [Arctic Shipping PR ](https://github.com/MPAS-Dev/geometric_features/pull/145).  Transect added in this PR is indicated in red at the bottom of this map:

E3SM_Foxe_Basin_Passages

and is also available in KML here:  

[Foxe_Basin_Through_Flow.kml.zip](https://github.com/MPAS-Dev/geometric_features/files/4901393/Foxe_Basin_Through_Flow.kml.zip)

Please include @milenaveneziani in the review list to check that this is what she was requesting.


proteanplanet:
@xylar Thanks for the fast turnaround. From my side, no need to see a WC14 mesh with this change, since it's almost open in #145 anyhow. It's also pretty obvious in EC30to60 what the impact would be. So I'm fine with proceeding to WC14 r03, which I believe is the last revision for E3SM V2, pending the approval of @milenaveneziani to this PR.

xylar:
## Testing

In the end, I felt like the time it would take to test this now and make sure it works as expected would potentially save quite a bit more hassle down the road.

### EC30to60
With Foxe Throughflow and Shipping Routes:
EC30to60

Without either:
old_EC30to60


### WC14to60
With Foxe Throughflow and Shipping Routes:
(Coming soon!)

Without either:
old_WC14

milenaveneziani:
oh I missed this PR entirely yesterday! Thanks @proteanplanet for putting this in and @xylar for testing.
Don't forget to rebase after the other one is merged in.

proteanplanet:
@xylar Thanks for testing.  With this, and #145, we should be ready to move forward with WC14 r03.

xylar:
@milenaveneziani, thanks for reviewing.  A rebase shouldn't be necessary.  That's convenient for looking at the diffs in the PR but makes no difference for the final state of the repo.

PR #147, opened 2020-08-09
Touch-Ups of Geographical Features for E3SM V2

proteanplanet:
This pull request is to be included in WC14r04 for E3SM V2 should we decide to shift to a new revision for other reasons, such as for changes in vertical resolution. This pull request is also to ensure that future mesh generation permits the included refinements for tidal model meshes during ICoM experimentation.  The reasons for the changes suggested in this pull request are threefold:  1) To ensure that boundary conditions for tides are as accurate as possible at low resolution for dissipative tests; 2)  To ensure that all U.S. states, including Hawaii, are represented in E3SM to the best of our ability; and 3) To make sure that highly recognizable geographic features, including Hawaii, the British Isles and the Mediterranean are equally as recognizable on E3SM and MPAS meshes for high level briefings and outreach purposes.  This pull request includes four elements:

a) An attempt to fix the English Channel, which for some reason is heavily widened in WC14r03 and previous meshes at the Pas de Calais as compared to the real world. The English_Channel_Deepen addition of @mark-petersen has been altered to include a higher resolution and slightly shallower transect (100m versus 120m).  Should this also prove inadequate, then there may be other problems with geometric features prescribed in the Channel.  The fix is annotated here:

English_Channel_Deepen_WC14r03_fix

b) A fix to the Inner Hebrides so that the Firth of Lorn appears at standard resolution (which it should):

Scotland_WC14r03_fix

c) Fix to Southern Italy so that Sicily, Salento, and Calabria are properly resolved at standard resolution:

Italy_WC14r03_fix

d) Fix to Hawaii to ensure that all of Hawai'i, Moloka'i, Maui, O'ahu, and Kaua'i appear on standard resolution maps, so that the state appears in Water Cycle simulations, but also so that sufficient boundary conditions exist in resolution tests on key amphidromic points in the Pacific during resolution degradation tests with tides in E3SM.

Hawaii_WC14r03_fix

This pull request has only a moderate priority, but should we choose to shift to WC14r04, the priority would elevate.  Parts (b)-(d) will most likely have a relatively straight-forward impact on the meshes.  However, part (a) will need to be carefully checked to understand if the change had any impact.


xylar:
## Mesh Results

### WC14 r04

Britain:
Britain

Hawaii:
Hawaii

Italy:
Italy

## EC30to60

Europe:
Europe

Hawaii:
Hawaii

@proteanplanet, I'll let you assess but my sense is that Italy, England and Scotland look good but Hawaii still needs work.  I wonder if making the shape at least for the Big Island more of a loop instead of a line would help.

xylar:
The results are on IC at:

WC14 r04:

```
/lustre/scratch4/turquoise/xylar/spin_up_WC14to60kmL60E3SMv2r04/ocean/global_ocean/WC14/init/initial_state
```

EC30to60:

```
/lustre/scratch4/turquoise/xylar/test_EC30to60_compass_0.1.9/ocean/global_ocean/EC60to30/init/initial_state
```

xylar:
To be honest, if Hawaii is a priority, the better way to handle it would be giving it at least a bit more resolution.

proteanplanet:
Thank you @xylar.  Your turnaround time is remarkable.  I'm on annual leave this week, but will try to sneak in a quick look at the mesh in the evening on Monday or Tuesday. Hawaii may be OK, so long as the islands fall within the 20m isobath, even if they don't reveal themselves.  However, I'm with you that a little resolution in this case may not hurt in this one special case, being a state.  The others (Italy, Britain) look fine.

vanroekel:
@xylar and @proteanplanet regarding Hawaii, I would pretty strongly prefer not increasing resolution around Hawaii.  In my view Hawaii is a pretty low priority for water cycle science questions, even though it is a state.

xylar:
> Hawaii may be OK, so long as the islands fall within the 20m isobath, even if they don't reveal themselves.

@proteanplanet, this is with the 60-layer vertical grid so the shallowest non-land bathymetry will be about 30 m deep.  Just keep that in mind.

> regarding Hawaii, I would pretty strongly prefer not increasing resolution around Hawaii. In my view Hawaii is a pretty low priority for water cycle science questions, even though it is a state.

@vanroekel, okay, noted.

vanroekel:
@mark-petersen yes I think it is fine to merge, but we should verify @proteanplanet doesn't want to tweak the shape file for Hawaii anymore.

proteanplanet:
@xylar, @mark-petersen, @vanroekel - Thanks for checking.  Please proceed, and thanks for taking this on.  I'm fine with all of this.  As far as Hawaii goes, the WC14 mesh is registering an improvement in bathymetry which is fine at this resolution. There are other applications (tides) where we will be increasing the resolution in these areas and using nearby long tidal gauge records, for which the Hawaii improvements will be most visible at higher resolution.

xylar (approved):

xylar:
I'm not going to make a new release of `geometric_features` or a new version of `compass` just for these changes unless someone feels strongly that it should be available right now.  It will come in with the next update on `compass`, which will likely be needed for other purposes.

PR #158, opened 2020-11-30
Add Arctic region aggregation for ocean and sea ice

xylar:
This merge adds aggregation functions for Arctic regions that are relevant to both ocean and sea-ice.

I also included clean-up in 2 places I discovered along the way, removing one irrelevant comment and fixing a typo.

### DIFF
--- a/docs/aggregation.rst
+++ b/docs/aggregation.rst
@@ -37,6 +37,15 @@ Meridional Overturning Circulation Basins
 The function :py:func:`geometric_features.aggregation.ocean.moc` aggregates
 basins for computing the MOC: Atlantic, IndoPacific, Pacific and Indian
 
+Arctic Regions
+~~~~~~~~~~~~~~
+
+The function :py:func:`geometric_features.aggregation.ocean.arctic` aggregates
+regions of the Arctic: Baffin Bay, Baltic Sea, Barents Sea, Canada Basin,
+Canadian Archipelago, Central Arctic, Chukchi Sea, East Siberian Sea,
+Greenland Sea, Hudson Bay, Irminger Sea, Kara Sea, Labrador Sea, Laptev Sea,
+North Sea, Norwegian Sea.
+
 Antarctic Regions
 ~~~~~~~~~~~~~~~~~
 
@@ -66,3 +75,17 @@ Barents Sea Opening, Bering Strait, Davis Strait, Drake Passage,
 Florida-Bahamas, Florida-Cuba, Fram Strait", Indonesian Throughflow,
 Lancaster Sound, Mona Passage", Mozambique Channel, Nares Strait, Tasmania-Ant
 Windward Passage.
+
+Aggregating Sea-ice Features
+----------------------------
+
+Arctic Regions
+~~~~~~~~~~~~~~
+
+The function :py:func:`geometric_features.aggregation.seaice.arctic` aggregates
+regions of the Arctic as defined by the
+`National Snow and Ice Data Center (NSIDC) <https://nsidc.org/>`_ that are
+relevant for sea ice: Baffin Bay NSIDC, Barents Sea, Beaufort Sea NSIDC,
+Canadian Archipelago NSIDC, Central Arctic NSIDC, Chukchi Sea NSIDC,
+East Siberian Sea NSIDC, Hudson Bay NSIDC, Kara Sea, Laptev Sea NSIDC.
+
--- a/docs/aggregation.rst
+++ b/docs/aggregation.rst
@@ -41,10 +41,11 @@ Arctic Regions
 ~~~~~~~~~~~~~~
 
 The function :py:func:`geometric_features.aggregation.ocean.arctic` aggregates
-regions of the Arctic: Baffin Bay, Baltic Sea, Barents Sea, Canada Basin,
-Canadian Archipelago, Central Arctic, Chukchi Sea, East Siberian Sea,
-Greenland Sea, Hudson Bay, Irminger Sea, Kara Sea, Labrador Sea, Laptev Sea,
-North Sea, Norwegian Sea.
+regions of the Arctic: Baffin Bay, Baltic Sea, Barents Sea, Beaufort Gyre,
+Beaufort Gyre Shelf, Canada Basin, Canadian Archipelago, Central Arctic, 
+Chukchi Sea, East Siberian Sea, Greenland Sea, Hudson Bay, Irminger Sea, 
+Kara Sea, Labrador Sea, Laptev Sea, North Sea, Norwegian Sea, and
+Arctic Ocean - no Barents, Kara Seas.
 
 Antarctic Regions
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


xylar:
@milenaveneziani, thanks for helping to clear things up in your comment (https://github.com/MPAS-Dev/MPAS-Analysis/pull/756#issuecomment-735606985).  It seems like the ``arcticRegions20200621.geojson`` file I pointed to was actually created with my incorrect definitions of the Arctic regions and should not be used.

Please check to make sure the aggregation functions I added here make sense.  They're really simple, so the main thing would be to make sure the list of regions they produce (see the modifications to the documentation) are what you expect.  Also, let me know if you're okay with having a `seaice` subdirectory under `aggregation` (even though the regions are currently in the `ocean`  folder in `geometric_data`).

docs/aggregation.rst
```diff
@@ -37,6 +37,15 @@ Meridional Overturning Circulation Basins
 The function :py:func:`geometric_features.aggregation.ocean.moc` aggregates
 basins for computing the MOC: Atlantic, IndoPacific, Pacific and Indian
 
+Arctic Regions
+~~~~~~~~~~~~~~
+
+The function :py:func:`geometric_features.aggregation.ocean.arctic` aggregates
+regions of the Arctic: Baffin Bay, Baltic Sea, Barents Sea, Canada Basin,
```
milenaveneziani, line 44: Above, you may want to modify to `of the Arcic (and subArctic)`.

milenaveneziani:
Thanks @xylar very much for doing this.

I just realized that, in my original `arcticRegions.geojson`, I also included the following regions:
1) `Beaufort Gyre`: subregion of the `Canada Basin` feature already included here, which is helpful when doing comparisons with observational data of freshwater content (Proshutinsky's papers).
2) `Beaufort Gyre shelf`: complements 1), as the region in 1) does not include the shelf.
3) `Arctic (no Barents/Kara Seas)`: helpful as some papers exclude the Barents and Kara Seas when talking about the Arctic. I think at one point you asked me about changing the name of this region, as MPAS-Tools was having trouble with spaces and slashes. I am totally fine with that. Could you please remind me what name you ended up using?

So, I think we could add these 3 regions. The first 2 have tag `Arctic_Proshutinsky`. The third one does not have a tag, but it is aggregated from the following regions: `Chukchi Sea`, `Canada Basin`, `East Siberian Sea`, `Laptev Sea`, `Central Arctic`.

docs/aggregation.rst
milenaveneziani: Could add: 'Arctic Ocean (no Barents/Kara Seas)' (or related name), 'Beaufort Gyre', and 'Beaufort Gyre shelf'.

docs/api.rst
```diff
@@ -95,3 +95,6 @@ Aggregation
    aggregation.ocean.antarctic
    aggregation.ocean.ice_shelves
    aggregation.ocean.transport
+
+   aggregation.seaice
+   aggregation.seaice.arctic
```
milenaveneziani, line 100: Nice. And yes, I like the idea of having the seaice folder with these regions in it.

geometric_features/aggregation/ocean/arctic_regions.py
```diff
@@ -0,0 +1,22 @@
...
+    -------
+    fc : geometric_features.FeatureCollection
+        The new feature collection with ocean-relevant Arctic regions
+    """
+    # Authors
+    # -------
+    # Milena Veneziani, Xylar Asay-Davis
+
+    fc = gf.read(componentName='ocean', objectType='region',
+                 tags=['Arctic'])
```
milenaveneziani, line 20: so basically here we would add the `Arctic_Proshutinsky` tag, and then add a few lines of code to aggregate the regions to make Arctic-noBarentsKaraSeas. I can do this if you like.

xylar:
@milenaveneziani, thanks, that clears up some more of my confusion, since I had noticed the missing features.  I'll add your suggested changes.

xylar:
The suggested name change was to `"Arctic Ocean without Barents and Kara Seas"`.  Something shorter would be nice but it can't have punctuation that isn't allowed in file names.

milenaveneziani:
how about 'Arctic-noBarentsKaraSeas'?

xylar:
That is more compact but I think the spaces do help.  I'm not sure if the dash is okay, probably so.

xylar:
Let's see if `Arctic Ocean - no Barents, Kara Seas` will work.

xylar:
The following:
``` python
from geometric_features.aggregation.ocean import arctic
arctic(GeometricFeatures()).to_geojson('arctic_ocean.geojson')
```
Produces a file with these names:
```
$ grep '"name": ' arctic_ocean.geojson 
                "name": "Baffin Bay",
                "name": "Baltic Sea",
                "name": "Barents Sea",
                "name": "Beaufort Gyre",
                "name": "Beaufort Gyre Shelf",
                "name": "Canada Basin",
                "name": "Canadian Archipelago",
                "name": "Central Arctic",
                "name": "Chukchi Sea",
                "name": "East Siberian Sea",
                "name": "Greenland Sea",
                "name": "Hudson Bay",
                "name": "Irminger Sea",
                "name": "Kara Sea",
                "name": "Labrador Sea",
                "name": "Laptev Sea",
                "name": "North Sea",
                "name": "Norwegian Sea",
                "name": "Arctic Ocean - no Barents, Kara Seas",
```
So far, so good.  I'm now going to rebuild the test package and test this out in MPAS-Analysis on Anvil.

xylar:
@milenaveneziani, if the results at:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/add_arctic_regions/arctic/ocean/index.html
look okay to you (given the very coarse resolution of the test simulation), I'll merge this and make another release so we can use this in MPAS-Analysis.  Let me know. There's no rush as far as I'm concerned, as long as you don't need this functionality in E3SM-Unified.

milenaveneziani:
@xylar: that looks great! It will be nice to have that functionality in MPAS-Analysis. Thanks.

docs/aggregation.rst
milenaveneziani: @xylar: don't forget to modify this text above before merging.

milenaveneziani (approved):

docs/aggregation.rst
```diff
@@ -37,6 +37,15 @@ Meridional Overturning Circulation Basins
...
+Canadian Archipelago, Central Arctic, Chukchi Sea, East Siberian Sea,
+Greenland Sea, Hudson Bay, Irminger Sea, Kara Sea, Labrador Sea, Laptev Sea,
+North Sea, Norwegian Sea.
```
xylar, line 47:
```suggestion
The function :py:func:`geometric_features.aggregation.ocean.arctic` aggregates
regions of the Arctic: Baffin Bay, Baltic Sea, Barents Sea, Beaufort Gyre,
Beaufort Gyre Shelf, Canada Basin, Canadian Archipelago, Central Arctic, 
Chukchi Sea, East Siberian Sea, Greenland Sea, Hudson Bay, Irminger Sea, 
Kara Sea, Labrador Sea, Laptev Sea, North Sea, Norwegian Sea, and
Arctic Ocean - no Barents, Kara Seas.
```

docs/aggregation.rst
xylar: Thanks, I missed these suggestions somehow.

PR #169, opened 2021-03-31
Add MOC Basins to aggregation by name

xylar:
The MOC basins are given a date stamp like other regions.  This will also make it easier to make masks for them in the same way we make other region masks.

### DIFF
--- a/geometric_features/aggregation/__init__.py
+++ b/geometric_features/aggregation/__init__.py
@@ -1,5 +1,6 @@
 from geometric_features.aggregation.ocean import basins, subbasins, \
-    antarctic, ice_shelves, ismip6, arctic as arctic_ocean, transport
+    antarctic, ice_shelves, ismip6, arctic as arctic_ocean, transport, \
+    moc
 from geometric_features.aggregation.seaice import arctic as arctic_seaice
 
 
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, do you just want to review this by inspection?  No need to do any testing.

milenaveneziani (approved):
sure. Looks good.

geometric_features/aggregation/__init__.py
```diff
@@ -52,6 +53,9 @@ def get_aggregator_by_name(region_group):
                'ISMIP6 Regions': {'prefix': 'ismip6Regions',
                                   'date': '20210201',
                                   'function': ismip6},
+               'MOC Basins': {'prefix': 'mocBasins',
+                              'date': '20210331',
```
xylar, line 57: Mask files without the southern transects will be named `<mesh>_mocBasins20210331.nc`.  Then, the ones with the southern transect (calculated via MPAS-Tools) will be `<mesh>_mocBasinsAndTransects20210331.nc`.  Does that seem right?

milenaveneziani: I wonder if we need both? just `_mocBasinsAndTransects` would be enough in my view, since these masks are only used in the MOC and MHT am's?

xylar: As a practical consideration, at the moment I don't have a good way of checking if `<mesh>_mocBasinsAndTransects20210331.nc` and not creating `<mesh>_mocBasins20210331.nc` if it does.  I'll have to think about how to do this without making things unnecessarily complicated.

xylar: Never mind, that's quite easy to handle.

xylar:
Tested with MPAS-Analysis on QU480 and QU240wLI meshes.  More testing will be performed after a release.

PR #170, opened 2021-04-04
Update transport transects date

xylar:
New transport transects were added, so new masks will need to be generated with a new `geometric_features` release.


xylar:
@milenaveneziani, I'm going to merge this when tests pass but wanted to ping you that this is happening.  When new transport transects get added, MPAS-Analysis will only work reliably if we also bump the date for transport transects so new masks get generated.

milenaveneziani:
Sounds good  @xylar. Thanks.

PR #174, opened 2021-05-02
Switch to spec file for development env.

xylar:
This allows us to use `conda create` instead of `conda env create`, which is more flexible and allows combining spec files from various repos.


xylar:
See https://github.com/MPAS-Dev/compass/issues/91#issuecomment-830702590 for discussion of this approach.

PR #178, opened 2022-04-01
Fix deprecation warning in MOC aggregator

xylar:
Shapely will soon stop allowing `MultiPolygon` objects to be treated as lists, so we want to use the `.geoms` property for this purpose instead.


xylar:
## Testing

I have run the QU240 mesh in compass through the `files_for_e3sm` test case, which uses the python masking.  With an `geometric_features` package built from this branch, I no longer see the deprecation warnings that I was seeing in previous runs.

PR #191, opened 2023-01-04
Fix importlib.resources import for python 3.8

xylar:


xylar:
## Testing

I ran pytest on an environment with python 3.8.  With the fix, I now see that test are passing and the only warning is internal to cartopy (nothing we can fix here).

PR #193, opened 2023-01-27
Remove seed points near islands

xylar:
This merge removes ocean seed points near Tierra Del Fuego and the Galapagos Islands.  These seed points are in danger of ending up as an isolated lake in many meshes, so it is not a safe seed point for flood filling the ocean to ensure connectivity.

The Tierra Del Fuego point has caused problems in MPAS-Analysis with the SO12to60 mesh (without ice-shelf cavities).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,11 @@ jobs:
       eval "$(conda shell.bash hook)"
       conda config --add channels conda-forge
       conda config --set channel_priority strict
-      conda install --yes conda conda-build mamba boa
+      conda update --yes --all
+      conda install --yes mamba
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I merged all the `seed_point` features into a single geojson file and plotted them:
seed_points_cyl
As expected, the 2 points with tags removed do not appear.

I ran the compass test case for making the SO12to60E2r5 mesh and initial condition.  As expected, there was as "lake" in Tierra Del Fuego without this fix but that lake disappeared with this fix.  I have added a check for no `maxLevelCell = 0` locations in this compass branch:

Which I plan to merge once this fix has gone in and a new `geometric_features` has been released.

PR #197, opened 2023-08-28
Add land blockage to entrance of Ilulissat Fjord (West Greenland)

milenaveneziani:
This adds a land blockage feature to the entrance of Ilulissat Fjord, in Disco Bay, West Greenland. 
We discovered that, at very high resolution (3 km and higher) and using the current GEBCO bathymetry version, a large bay/lake appears in the middle of Greenland. This is because GEBCO has the Greenland bedrock topography, rather than the ice elevation, and therefore at high enough resolution the fjord becomes a channel to the inner glacier..

This blockage should hopefully prevent all this in the future.


stephenprice:
That makes sense -- the connection to the ocean is only at the narrow fjord for the Jakobshavn outlet glacier. The opening would not be resolvable until we get down to a few km in resolution.

milenaveneziani:
Thanks @stephenprice for checking. Do you think the blockage feature is long enough to cover the entrance and include Jakobshavn glacier?

@xylar: not sure what is up with the failed checks, but let me know if something in the files needs to be adjusted.

xylar:
@milenaveneziani, the failed checks are unrelated.  I'll just remove them.

stephenprice:
@milenaveneziani -- Yes, I think this will be more than sufficient. It could probably be shorter and moved further back (to the east), but I think this will suffice and I don't think there would be any harm putting it here for now. I'm not entirely sure what other side effects of these blockages are though, so @xylar may also want to weigh in.

xylar:
@stephenprice and @milenaveneziani, is there any risk of a similar leakage from Petermann Glacier?  It didn't look like it from this mesh but maybe at 2 km or higher?

xylar:
This blockage is good.  Leave it as it is.

milenaveneziani:
The only side effect is that we loose the fjord all together: there will be land behind the blockage. So, if you think we want to retain that, I can certainly push the transect inland.

xylar:
Can we add this, too, just in case?

``` json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            -61.99344324352276,
            80.83404595432123
          ],
          [
            -60.20449437539784,
            80.96559467786543
          ]
        ],
        "type": "LineString"
      }
    }
  ]
}
```

stephenprice:
Looking at a detailed map of bed topography on my wall ... I don't think Petermann poses the same risk. Those little tributaries reaching up north that you can see in the fig above seem to die out because of some highlands that are only bisected by a very very narrow and isolated trough.  Humboldt might actually be more of a risk. I'll try to post a fig. of the bed topo here in a minute ...

xylar:
Here, if you want to keep the fjord, you could do:
``` json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            -50.017306316678116,
            69.24164784625944
          ],
          [
            -49.586940141736136,
            69.27126272249464
          ],
          [
            -49.45905990689619,
            69.1026172201299
          ],
          [
            -49.94844772868825,
            69.08769977505611
          ]
        ],
        "type": "LineString"
      }
    }
  ]
}
```

xylar:
Okay, maybe ignore Petermann...

milenaveneziani:
Here is a picture I found in a GRL paper (https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2017GL074954):

image

stephenprice:
zoom in on bedmachine 4. arrows show points that I think would block any continuous connection to the ocean
Screenshot 2023-08-28 at 1 09 08 PM

trhille:
I don't understand the relationship between those two panels. From the right-hand panel it really looks like the Petermann and Humboldt fjords connect to the interior (as I've always thought they did), while the left-hand panel makes it look like there are bed features that block that connection.

milenaveneziani:
yes, I agree that those two panels do not agree. Maybe Humboldt's connection is even narrower and the first panel doesn't show it?
In any case, I will just add a double blockage as Steve suggested.

milenaveneziani:
What do you think about these two:

Screen Shot 2023-08-28 at 1 52 13 PM
Screen Shot 2023-08-28 at 1 52 43 PM

xylar:
Both look great to me.

xylar:
>  I see those end of file red checks again, not sure where they come from.

Don't worry too much about those.  Ideally, you would have a blank line at the end of the file (depending on your editor) but it doesn't really matter.

milenaveneziani:
I see. Good to know.

Thanks for the quick feedback everyone.

xylar:
I think the tool that extracts the features doesn't do that. so don't worry about it.

It looks like my commit to remove the OSX tests got removed in a rebase.

xylar:
I'll merge this once the tests pass and make a release (and a PR to update compass) in the morning.

milenaveneziani:
yes, sorry, I forgot to incorporate it before rebasing.

xylar (approved):
@milenaveneziani, always a pleasure to have another expert working on geometric_features!

milenaveneziani:
@xylar: while chatting with Gennaro about this, I have been wondering: is it possible that this issue will crop up for Antarctica as well (especially EA/Peninsula/Weddell area) when we start venturing in 3km resolution grounds?

xylar:
@milenaveneziani, no, we use a different dataset for Antarctica -- BedMachine Antarctica v3.  We explicitly have topography for the bathymetry and floating ice shelves, and we use the grounding line from that dataset (interpolated to the MPAS mesh) as the land mask for culling.  So there is no risk of this.  Presumably, the right solution for Greenland would be to do the same with BedMachine Greenland, but I haven't had the need to do this yet.

PR #199, opened 2024-05-10
Add ISMIP6 Greenland regions

cbegeman:
Add regions around the Greenland continental shelf as defined by ISMIP6.


cbegeman:
MPAS-Analysis results using these regions can be found here: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.cbegeman/analysis/20231105.v3b01.piControl.chrysalis/gris_regions_0001-0050_climo_0001-0050/ocean/index.html

cbegeman:
@xylar Do you have thoughts on whether the existing `ISMIP6 Regions` group should have the Greenland ISMIP6 regions added? My inclination was no.

xylar:
@cbegeman, it seems like we could consider renaming the `ISMIP6 Regions` to `ISMIP6 Antarctic Regions`.  I wouldn't try to have the Antarctic and Greenland regions in the same group.

xylar:
Looking around, the `ISMIP6 Regions` are only used in 2 places right now.  One is just a utility script so not something I feel a strong need to maintain but which would be easy to update:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/e57f81c0474674fac5dbaeb4eecbcb438f42ec84/utility_scripts/make_ismip6_region_masks.py#L14
And the other is in Compass:
https://github.com/MPAS-Dev/compass/blob/74edf148186599dbcfe39f738af4f59bb0782e80/compass/ocean/tests/global_ocean/files_for_e3sm/diagnostic_masks.py#L86

Both of these would be quick to update when you add the Greenland Regions.

xylar (approved):

alexolinhager (commented):
@cbegeman, thank you for putting this all together! Looks great, and no changes to request

alexolinhager (approved):

PR #202, opened 2024-05-23
Add Antarctic isobath at -1000.0 m

xylar:
This transect will hopefully be useful for studying export of water masses.


xylar:
@alicebarthel, this is the geojson transect I was suggesting you might want to use for studying water-mass transport.  The next steps would be to use it to get a list of edges and distances along the transect.  I will look into that next.

xylar:
@alicebarthel, okay, so it's more complicated than I was hoping.  We have tools for:
* getting edge transects and the edge sign (https://mpas-dev.github.io/MPAS-Tools/stable/mesh_conversion.html#mask-creation-with-python-multiprocessing)
* getting distance along the transect as well as info for plotting a section of a 3D field at cell centers (e.g. https://mpas-dev.github.io/MPAS-Tools/stable/ocean/visualization.html)
But not a tool for getting the distance along the transect together with the edges and edge signs.

I'm going to have to think about this.

feature_creation_scripts/bedmachine_antarctica_v3_1000m_transect/bedmachine_antarctica_v3_1000m_transect.py
```diff
@@ -0,0 +1,138 @@
...
+    width = 30e3
+    x0 = x[0]
+    y0 = y[0]
+    square = shapely.geometry.Polygon([(x0 - width, y0 - width),
+                                       (x0 + width, y0 - width),
+                                       (x0 + width, y0 + width),
+                                       (x0 - width, y0 + width),
+                                       (x0 - width, y0 - width)])
+
+    difference = transect.difference(square)
```
xylar, line 77: I ended up needing to cut a little square out of the transect.  The mask creator (specifically the code that finds edge signs) was totally confused by the transect being a closed loop.  We could make the square less than 60 km in size if we want to.

Here's what that gap ends up looking like:
edge_sign

milenaveneziani:
@xylar pointed me to this from the MPAS-quickviz PR and couldn't help taking a look :).
I'm a bit concerned by the intersection of the 1000m isobath with the coastline between ~20W and ~20E:
Screen Shot 2024-06-05 at 4 22 38 PM
If I look at a plot of bed topography and grounding line, it seems like the 1000m isobath should follow a more offshore route in the eastern Weddell:
Antarctic-BEDMAP2-topography-Fretwell-et-al-2013-rebounded-after-the-removal-of

xylar:
@milenaveneziani, I agree that looks odd but this contour is straight from BedMachine v3 so I think it will be consistent with E3SM v3 meshes. We could make a Bedmap 2 version for v2 meshes but I'd prefer not to if it can be avoided.

milenaveneziani:
I understand. I worry that the region might come out 'interrupted', if you know what I mean. Maybe Alice could try to compute a regional mask for one of the v3 meshes and check that with paraview?

xylar:
@milenaveneziani, another thing to keep in mind is that the ice-shelves sometimes overhang the 1000m isobath in Dronning Maud Land.  So in meshes with ice-shelf cavities, the coastline that you're seeing plotted is the calving front, not the grounding line and the transect should be unbroken.  Working with the E3SM v2 SORRM mesh, the transect seems to be fine (it's continuous).  I haven't tried the E3SM v3 Icos30 or the E3SM v2 EC30to60 meshes yet.

xylar:
@alicebarthel, I would like to make a new geometric_features release with this feature added.  Do you have time to review it today or tomorrow?

PR #203, opened 2024-07-18
Add Carroll Inlet as a critical passage

xylar:
This feature keeps access to Carroll Inlet adjacent to Stange Ice Shelf in the Bellingshausen Sea region of Antarctica.

image
image

### DIFF
--- a/geometric_data/ocean/transect/Carroll_Inlet/transect.geojson
+++ b/geometric_data/ocean/transect/Carroll_Inlet/transect.geojson
@@ -1,0 +1,33 @@
+{
+    "type": "FeatureCollection",
+    "features": [
+        {
+            "type": "Feature",
+            "properties": {
+                "name": "Carroll Inlet",
+                "tags": "Critical_Passage",
+                "object": "transect",
+                "component": "ocean",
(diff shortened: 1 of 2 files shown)


xylar:
Here's the image I used to identify the name of the inlet:
https://upload.wikimedia.org/wikipedia/commons/3/38/Bryan_Coast_to_Ellsworth_Land.jpg

darincomeau (approved):
Agree we want this open to avoid a land(-ice) locked ocean region - thanks @xylar !

xylar:
My testing with this didn't work, but I found that the reason is that critical passages open up the ocean but don't change the `landIceFration` so the result is that a critical passage just changes grounded land-ice to floating land ice.  I'll need to fix that in Compass.

xylar:
In my testing in https://github.com/MPAS-Dev/compass/pull/846, this seems to be accomplishing what we want it to.  If @darincomeau agrees, I'll merge both this PR and that one, and get to work on https://github.com/MPAS-Dev/compass/pull/845

darincomeau:
Agree!

PR #205, opened 2024-09-02
Add Southern Hemisphere region

cbegeman:
Add a region for the Southern Hemisphere


cbegeman:
@xylar Can you take a look at this and let me know if you see any issues? I'm having issues using it with MPAS-Analysis but I think something might be off with my editable environment because when I use a modified pytest to check these features they seem fine.

xylar:
@cbegeman, yes, it looks fine to me.

cbegeman:
@xylar Would you be willing to give this a try? I was changing the `regions` option in `timeSeriesOHCAnomaly` but I don't know if I'm getting the syntax right on the region name that corresponds to Southern Hemisphere. It seemed to fail with or without an underscore.

xylar:
@cbegeman, sure, I can give it a try tomorrow.

xylar:
@cbegeman, I see the problem.  `timeSeriesOHCAnomaly` uses output from the online [Regional Stats](https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/analysis_members/mpas_ocn_regional_stats.F) analysis member.  This is really unfortunate because those regions are hard-coded.

We could presumably write a post-processed version of OHC that could take a region from a mask but that doesn't yet exist.  Or alternatively wouldn't OHC be trivial to calculate from mean temperature in a region (using `timeSeriesOceanRegions`), volume of the region, heat capacity, etc.?  Or am I missing something?

milenaveneziani:
@xylar: you are not missing anything. I always thought that OHC analysis tasks could be done in post-process mode (along with regional statistics, to be honest). In fact, I believe that the other OHC task, the climatological maps, is done using the layer-weighted averaged temperature.

cbegeman:
Thanks @xylar and @milenaveneziani. It seems like post-processing OHC makes a lot of sense and could allow the AM to be removed in OMEGA. I'll implement it in MPAS-Analysis. This work was motivated by being able to compare our SORRM OHC with published values from CMIP6 models.

geometric_features/aggregation/__init__.py
```diff
@@ -43,7 +43,7 @@ def get_aggregator_by_name(region_group):
                                           'date': '20201130',
                                           'function': arctic_seaice},
                'Ocean Basins': {'prefix': 'oceanBasins',
-                                'date': '20200621',
+                                'date': '20240830',
```
xylar, line 46: @cbegeman, I updated the date stamp for Ocean Basins.  I hope you don't mind.

This is needed so new region masks get created in MPAS-Analysis and Compass that include Southern Hemisphere.

cbegeman:
@xylar Does this look ready for review to you? Seems like you successfully used it in your MPAS-A PR

xylar (approved):
This looks good to me and, yes, I was able to test it in MPAS-Analysis.  It successfully produced ocean heat content integrated over the region:
https://github.com/MPAS-Dev/MPAS-Analysis/pull/1034#issuecomment-2455218915

xylar:
@cbegeman, remind me that we'll want to cache some mask files for typical meshes with these new regions to save MPAS-Analysis from generating them every time it runs.  This would be once this is being used in MPAS-Analysis, of course.

cbegeman:
Thanks @xylar for your testing. I'll merge now.

cbegeman:
@xylar Actually, I don't have permissions to merge so you'll have to do it.

cbegeman:
@xylar Thanks!

PR #206, opened 2024-09-10
Update supported python versions

altheaden:
Since conda-forge has stopped supporting python 3.8 and started supporting python 3.12, I am updating the version numbers in various places to reflect that change.


xylar (approved):
@altheaden, looks great!

PR #210, opened 2024-10-29
Extend NASA GrIS regions into ocean

alexolinhager:
This PR introduces manually modified versions of the NASA GrIS in `geometric_data/landice/region` (`eastCentralGreenland1`, etc.), so that each region now extends into the ocean to the continental shelf. The boundaries of some regions were also modified so one region flows into each fjord. For now, the PR adds the newly modified regions without replacing the originals in case existing analyses are dependent on these regions.


alexolinhager:
Thanks @cbegeman, that's a good suggestion. I just pushed a commit with a new naming convention

xylar:
@alexolinhager, I see that your features aren't using the geojson formatting that is standard here, you're using a more compact format that isn't as easy to read.  Could you follow the second procedure here to merge them all into a single feature collection and then split them back out into individual features to clean things up?
https://mpas-dev.github.io/geometric_features/main/adding_features.html

xylar:
The important bit is:
``` python
# split the feature collection back into individual features within
# ./geometric_data to clean things up
gf.split(fc)
```
It looks like you have the tags in the database and all that.

docs/aggregation.rst
```diff
@@ -88,16 +90,26 @@ Amundsen Sea Deep, Eastern Ross Sea Shelf, Eastern Ross Sea Deep,
...
 ISMIP6 Greenland North Shelf, ISMIP6 Greenland North West Shelf,
 ISMIP6 Greenland South East Shelf, ISMIP6 Greenland South West Shelf.
 
+NASA Greenland Regions
+~~~~~~~~~~~~~~~~~
+
+The function :py:func:`geometric_features.aggregation.landice.nasa_greenland`
+aggregates regions of Greenland drainage basins taken from NASA: 
+(https://earth.gsfc.nasa.gov/cryo/data/polar-altimetry/antarctic-and-greenland-drainage-systems)
+The drainage basins have been altered to extend to the continental shelf. Boundaries between basins
```
matthewhoffman, line 109: Is this suggestion correct?
```suggestion
The drainage basins have been altered to extend to the continental shelf break. Boundaries between basins
```

docs/aggregation.rst
```diff
@@ -88,16 +90,26 @@ Amundsen Sea Deep, Eastern Ross Sea Shelf, Eastern Ross Sea Deep,
...
 
-The function :py:func:`geometric_features.aggregation.ocean.greenland`
+The function :py:func:`geometric_features.aggregation.ocean.ismip6_greenland`
 aggregates regions of the continental shelves around Greenland based on the
 ISMIP6 regions: ISMIP6 Greenland Central East Shelf,
 ISMIP6 Greenland Central West Shelf, ISMIP6 Greenland North East Shelf,
 ISMIP6 Greenland North Shelf, ISMIP6 Greenland North West Shelf,
 ISMIP6 Greenland South East Shelf, ISMIP6 Greenland South West Shelf.
 
+NASA Greenland Regions
```
matthewhoffman, line 103: Can you update the descriptions for "NASA Greenland Regions" and "ISMIP6 Greenland Regions" to indicate if each is intended for use by ocean or ice sheet applications or both?

geometric_features/aggregation/__init__.py
```diff
@@ -51,9 +51,12 @@ def get_aggregator_by_name(region_group):
                'Ocean Subbasins': {'prefix': 'oceanSubbasins',
                                    'date': '20201123',
                                    'function': subbasins},
-               'Greenland Regions': {'prefix': 'greenlandRegions',
+               'ISMIP6 Greenland Regions': {'prefix': 'ISMIP6GreenlandRegions',
```
matthewhoffman, line 54: @xylar , do you prefer that the `prefix` for this and the NASA Greenland Regions to use lower/camel case for the acronyms like is done for the "ismip6Regions" prefix?

geometric_features/aggregation/landice/nasa_greenland_extended.py
```diff
@@ -0,0 +1,41 @@
+def nasa_greenland(gf):
+    """
+    Aggregate NASA Greenland ice basin region extended to continental shelf.
```
matthewhoffman, line 3:
```suggestion
    Aggregate NASA Greenland ice sheet basin regions extended to continental shelf.
```

matthewhoffman (changes requested):
@alexolinhager , this looks good.  I just have a few minor suggestions.

I loaded all the new regions (green) into QGIS and overlaid the GIS outline (pink) and one of the ISMIP6 ocean regions (red) for comparison.  The background is BedMachine bed topography colored between -2000 and 0 m.  Everything looks as expected!  This will be nice to have available.


image

docs/aggregation.rst
alexolinhager: Yes that's more descriptive

geometric_features/aggregation/__init__.py
alexolinhager: It makes sense to me just to change the caps to make this consistent

matthewhoffman (approved):

PR #212, opened 2024-11-04
Whitelist all sphinx-multiversion versions

altheaden:
This PR just removes an old fix where we were blacklisting sphinx-multiversion version 1.7.2. We no longer need to do so, so I've removed that bit. I also fixed a small typo in the same file.


xylar:
@altheaden, this was actually just an issue where we copied something over from MPAS-Analysis by mistake.  It blacklists 1.7.2, which has docs that cannot be built.  But that obviously doesn't apply `geometric_features`, which doesn't have such a version.

xylar (approved):

xylar:
The fix looks great!

PR #213, opened 2024-11-04
Add python 3.13

altheaden:
This PR adds support for python 3.13. No changes were needed for this update beyond adding the version number to `setup.py` and the CI files. I also updated the README to suggest python 3.13 as we did when we added python 3.12.


ci/python3.13.yaml
```diff
@@ -0,0 +1,8 @@
+channel_sources:
+- conda-forge,defaults
+pin_run_as_build:
+  python:
+    min_pin: x.x
+    max_pin: x.x
+python:
+- 3.13.* *_cpython
```
xylar, line 8: For complicated reasons, they decided to name this differently for python 3.13:
```suggestion
- 3.13.* *_cp313
```

xylar (approved):
@altheaden, I went ahead and made a small fix.  I don't think we're using the yaml files in ci for anything anyway but they could come up if someone wanted to build the conda package locally.

With that fix, things look great!

altheaden:
Ok, thanks!

xylar:
I don't know why CI passed here but then gave us trouble afterward.  Hopefully, #214 fixes it.

PR #216, opened 2024-12-09
Add CI workflow to automatically update pre-commit and its dependencies

altheaden:
This adds a GitHub actions workflow that will run the `pre-commit autoupdate` command to update `pre-commit` and its dependencies. The workflow is set to run weekly on Wednesdays and should automatically create a pull request with the updates when they are available. 

I also included a small fix for the build workflow, which will hopefully stop failed tests from canceling other tests (to isolate bugs associated with a particular python version, for instance).

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -26,6 +26,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+      fail-fast: false
     steps:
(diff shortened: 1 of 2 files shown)


altheaden:
## Testing
I didn't do any additional testing for this workflow apart from what I've done in `mache` and `pyremap`. On Wednesday we'll see how this workflow runs, which will hopefully be just as well as it did for `mache`.

xylar (approved):
@altheaden, indentation looks good!  As does everything else.

PR #217, opened 2024-12-19
Remove unused tag

alexolinhager:
Removes bug in `geometric_features/features_and_tags.json`, in which an unused region was listed. `northGreenland4_oceanExtended` was absorbed by its neighboring regions when extending into ocean, so should no longer be listed in this file.

### DIFF
--- a/geometric_features/features_and_tags.json
+++ b/geometric_features/features_and_tags.json
@@ -890,9 +890,6 @@
       "northGreenland3_oceanExtended": [
         "northGreenland"
       ],
-      "northGreenland4_oceanExtended": [
-        "northGreenland"
-      ],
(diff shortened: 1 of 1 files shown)


xylar (approved):

xylar:
Indeed, it is obvious that this feature no longer exists:
https://github.com/MPAS-Dev/geometric_features/tree/main/geometric_data/landice/region
Presumably, we should also have a sanity check to make sure all the features in the json file actually exist.  I'll make a separate PR for that.

xylar:
As mentioned in #218, #219 and #220, I think there are more things that need to be fixed.  I think it might make sense to close this PR and let #220 take care of this fix along with several others.

PR #218, opened 2024-12-20
Fix sea-ice historical median features

xylar:
The name of the feature needs to be exactly the same as the folder it's in (except ' ' --> '_') and the name of the component needs to be `seaice` instead of `sea ice`.


xylar:
Approved by @alexolinhager in #220

PR #220, opened 2024-12-20
Add test to make sure `features_and_tags.json` looks as expected

xylar:
This test generates a new `features_and_tags.json` based on the contents of `geometric_data`.

The contents of `geometric_features/features_and_tags.json` have been updated with the correct data produced with:
`geometric_features.utils.write_feature_names_and_tags()`.

PR #221, opened 2024-12-20
Add `pre-commit` for code linting

altheaden:
This PR adds `pre-commit` for automatic code linting. I've also added `pre-commit` to the CI build workflow to automatically lint any incoming PRs.

I'm including changes that will satisfy all `pre-commit` hooks except `flake8`. The changes `flake8` wants require manual editing to many code files, so we're opting to only deal with those issues on a file-by-file basis as new changes come in. Because of this, the linting step of CI for this PR will fail, but that's expected.

Finally, I'm including a small bugfix to the docs CI workflow.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,44 @@
...
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+
+  # Can run individually with `pre-commit run isort --all-files`
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+        exclude: "geometric_features/__init__.py"
```
altheaden, line 17: @xylar isort reordering the imports for this file broke everything, so I've excluded it here.

xylar: Let's follow up on that in a separate PR.  We don't need all the `__future__` stuff anymore.  That was for python 2.  Maybe that will help. But if that's not the problem, we maybe need to add missing imports somewhere else.

geometric_features/features_and_tags.json
```diff
@@ -1928,11 +1928,11 @@
       "August Historical Median Sea Ice Extent":[
         "QGreenland",
         "Historical_Sea_Ice_Extent"
-      ], 
+      ],
```
xylar, line 1931: Let's exclude this file from pre-commit checks because I just added a test in #220 to make sure it's formatted as expected and the two might slightly disagree.  In particular, the `json` output seems to not put the desired newline at the end of the file.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,44 @@
+exclude: "docs|.git|geometric_data"
```
xylar, line 1:
```suggestion
exclude: "docs|.git|geometric_data|geometric_features/features_and_tags.json"
```

xylar (commented):
We need to add `pre-commit`, `isort`, etc. here as well:
https://github.com/MPAS-Dev/geometric_features/blob/main/dev-spec.txt#L14-L17

geometric_features/features_and_tags.json
xylar: I went ahead and took care of this in order to get this in.  I hope that's okay.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,44 @@
...
+        additional_dependencies: [flake8-isort]
+
+  # Can run individually with `pre-commit run mypy --all-files`
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.13.0
+    hooks:
+      - id: mypy
+        args: ["--config=pyproject.toml", "--show-error-codes"]
+        verbose: true
+        additional_dependencies: ['types-requests', 'types-PyYAML']
```
xylar, line 44: I think we had talked about this in our meeting but I would rather leave `mypy` out on this repo.  In my experience, it's a step too far unless you start with it from the beginning.

xylar:
We're purposefully ignoring the `flake8` failures here for now.  So I will merge without CI passing for now.

xylar (approved):

xylar:
Thanks @altheaden!!

PR #222, opened 2024-12-21
Clean-up on imports and pre-commit

xylar:
This merge removes and very old `__future__` imports needed for python 2 support.

It also makes changes needed to allow `isort` to include `geometric_feature/__init__.py`:
* moves version info to `version.py`
* updates `__main__.py` to not use shortcut in `__init__.py` and to use version info from `version.py`.
* sorts imports in `__init__.py`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
     rev: 5.13.2
     hooks:
       - id: isort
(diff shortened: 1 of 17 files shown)


xylar:
As in #221, I'm going to ignore flake8 errors for now because we don't want to deal with them all in one go.

PR #223, opened 2024-12-21
Fix `flake8` lint in `geometric_features`

xylar:
This merge addresses all the lint identified by `flake8` within the `geometric_features` package.  Notably, it doesn't address the `feature_creation_scripts` directory, which is there for provenance and we don't necessarily try to keep it clean or up-to-date.


xylar:
@altheaden, it seems like `flake8` and maybe the other pre-commit features are running on all files in CI and not just the modified files.  Can you figure out how to change that?  Feel free to push changes here.

xylar:
@altheaden, as I just slacked you, I got it.

PR #226, opened 2025-01-07
Switch setup configurations to `pyproject.toml`

altheaden:
In this PR I've switched the setup configurations from the `setup.py` approach to the more updated `pyproject.toml` approach that we're adopting in other repos.


pyproject.toml
```diff
@@ -0,0 +1,85 @@
...
+warn_unused_ignores = true
+warn_redundant_casts = true
+warn_unused_configs = true
+
+[build-system]
+requires = ["setuptools>=60"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.packages.find]
+exclude = ["docs*", "examples", "geometric_data", "contributors"]
```
xylar, line 66:
```suggestion
include=["geometric_features", "geometric_features.*"]
```

xylar (commented):
Yep, other than the one switch from `exclude` to `include` that we discussed yesterday, I think this looks great and is ready to merge.  I'll merge once that change is in and once CI has passed.

I'm going to go ahead and commit the change.  I hope that's okay with you.

xylar (approved):

PR #227, opened 2025-01-17
Use separate conda cache directories in CI

altheaden:
This PR updates the build workflow to use separate directories for each python version for conda caching. Issues with this seem to have been crashing CI runs lately.

I also set the pre-commit autoupdate workflow to use the same `PYTHON_VERSION` convention as the docs workflow, for consistency.


xylar (approved):
@altheaden, again, looks just like what we want!

PR #234, opened 2025-04-01
Switch linting from `flake8` and `isort` to `ruff`

altheaden:
This PR replaces linting with `flake8` and `isort` with `ruff`.

I did not lint all files, except to add implicit re-exporting to each of the `__init__` files which should fix all of the `F401` errors. Any additional linting will be done on a file-by-file basis as files are edited and updated.


xylar (approved):
@altheaden, great work as always!

PR #264, opened 2025-09-19
Drop support for Python 3.9

altheaden:
This PR drops support for Python 3.9 and removes various places where we test with different python versions, such as in CI.

### DIFF
--- a/dev-spec.txt
+++ b/dev-spec.txt
@@ -2,7 +2,6 @@
 # $ conda create --name <env> --file <this file>
 
 # Base
-python>=3.9
 cartopy
 matplotlib-base
 numpy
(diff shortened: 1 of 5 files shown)


dev-spec.txt
```diff
@@ -2,7 +2,6 @@
 # $ conda create --name <env> --file <this file>
 
 # Base
-python>=3.9
```
xylar, line 5: This needs to be kept but set to `python >=3.10`.

xylar (commented):
@altheaden, this is great!  Just one thing that needs fixing before I can merge it.

xylar (approved):

xylar:
@altheaden , thanks! Looks good.

PR #299, opened 2026-07-25
Add AIS & GIS flowline transects

matthewhoffman:
This merge adds landice flowline transects for major outlet glacier and ice streams of the Antarctic Ice Sheet and Greenland Ice Sheet.  They can be used for evaluating MALI input and output states.  

The AIS flowlines (71) originated in the [sl-ice](https://github.com/yuwang115/sl-ice) repository and were converted to geomtric_features format with some additional cleanup.  The flowlines were also extended 40km beyond their original termination at the calving front to allow sampling of MALI output for states where the calving front extends further.  Where multiple flowlines for the same glacier existed, only one was retained.

The GIS flowlines (277) originated from this paper and archive:
    https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2020GL090112
    https://zenodo.org/records/4284759
Like the AIS transects, they were converted to geometric_features format and standardized.  They were also extended by 40 km, but this had mixed results for the GIS glaciers as many of the extensions do not follow the fjord centerline or even stay in the ocean.  Still, having some extension is better than nothing.  They could be cleaned up in the future as needed.  Where multiple flowlines for the same glacier existed, only one was retained.  Official and common (foreign) names are matched from the listing of https://tc.copernicus.org/articles/9/2215/2015/.


matthewhoffman:
The individual transects can be inspected by going to the list of commits, browsing the repo at the most current commit, and then navigating to the geojson files added where github with plot them on a map.  I have pasted a few example maps here that were generated by my conversion scripts.  I have plots for all transects and I can share if desired.

## AIS

image

image

image

image

image

## GIS

(Note that some extensions are good, some are so-so, and some are terrible.)

image

image

image

image

image

image

image

matthewhoffman:
@xylar , I'm only noticing now that this changes 1 million lines of code and adds 35 MB of geojson files to the repo.  I'm assuming that's ok, but if that's too big, I can adjust my processing to coarsen the transects.

xylar:
@matthewhoffman, that's pretty normal for this repo.  Please proceed.

matthewhoffman:
@trhille , I realized I should add these maps showing all of the transects at once.   

For your consideration: I'm undecided if we should introduce these transects as-is, with their full length from ice divide to beyond the margin, or if we should clip them on the inland side to something like 100 km upstream of the GL.  The decision here may depend on how we want to use them.  My plan is to use them for at least 2 things: evaluating calving front (and maybe GL) position for the most important point laterally on each glacier (e.g. as an alternative calving calibration metric to area change); plotting velocity and thinning profiles in post-processing.  For both of these purposes the full transect is overkill, but I'm still inclined to keep them as full transects for now for completeness until we've had a chance to work with them.

antarctic_transects_overview


greenland_transects_overview

PR #300, opened 2026-07-28
Add AIS calving calibration regions

matthewhoffman:
This merge introduces 28 landice regions for Antarctica that divide the ice sheet into regions appropriate for calibrating and analyzing MALI calving behavior.  The regions originated from the existing iceshelf/regions but modified them in a few ways: 
* many small ice shelves were combined into aggregate regions
* regions that were subsections of a single ice shelf (e.g. Ross, FRIS) were combined into a single region for that whole ice shelf.
* regional aggregation was established by first defining major ice shelves and glaciers and having them be their own regions.  Then, remaining small ice shelf regions were grouped into logical groupings around the remaining perimeter of the ice sheet.  The regional granularity is meant to be such that each region could take on different values of calving parameter(s) and produce good AIS-wide behavior.
* Some of the regions in East Antarctica did not extend far enough seaward to capture the tip of ice shelves on the MALI mesh, or were dangerously close the to ice margin during calibration experiments where the modeled extent might grow forward.  This was because the original regions were defined by the continental shelf break, which is very close to the continent across a lot of East Antarctica.  For the purposes of ice-shelf calving calibration, that outer boundary isn't relevant, so I extended all EA regions about 200km northward.  WA regions were left unchanged.

Regions are given names `geometric_data/landice/region/ice_shelf_extended_*`.  They are also give the tag `ais_ice_shelf_calving_calibration`.


matthewhoffman:
This PR is sort of a companion to #299.  It defines regions over which calving calibration can be evaluated in MALI.  The existing ice-shelf regions were to granular and some were not extended far enough.  Existing ISMIP and IMBIE regions were not granular enough.  So this set of features is specific to the purpose of evaluating calving in MALI and defining a level of granularity over which we want calving parameter to be constant.  

@xylar , I made them all have a certain naming convention and gave them a tag.  Feel free to suggest a different way of handling their organization (e.g. putting them in their own high level directory instead of 'landice').

Here are the aggregated regions:
image


And here are the extended EA regions:
image

I have a set of ad hoc scripts to do the aggregation and extension in a separate branch on my fork at https://github.com/matthewhoffman/geometric_features/tree/calv-calib-generation-tools.  I didn't want to pollute this repo by including them, but I want to keep the around for potential tweaks to these regions.  Keeping them perpetually in that branch on my fork seems sufficient for that.

xylar:
@matthewhoffman, this is fine.  If you want, you could define a more "formal" aggregator with a date stamp like we do for ocean and sea-ice regions:
https://github.com/MPAS-Dev/geometric_features/blob/main/geometric_features/aggregation/__init__.py

xylar:
If you can include the creation script for provenance, that's always good:
https://github.com/MPAS-Dev/geometric_features/tree/main/feature_creation_scripts

Also, I am listed as the author of the regions. Maybe it should be you.

